PR #23, opened 2016-08-09
Adding tests for dirty state type modifiers.

gallandarakhneorg:
see eclipse/xtext-eclipse#28
see eclipse/xtext-eclipse#30

Signed-off-by: Stéphane Galland galland@arakhne.org

### DIFF
--- a/org.eclipse.xtend.core.tests/META-INF/MANIFEST.MF
+++ b/org.eclipse.xtend.core.tests/META-INF/MANIFEST.MF
@@ -20,7 +20,8 @@ Require-Bundle: org.eclipse.xtend.core,
  org.objectweb.asm;bundle-version="5.0.1",
  org.eclipse.jdt.core,
  org.eclipse.xtext.builder.standalone,
- org.eclipse.xtext.java
+ org.eclipse.xtext.java,
+ org.eclipse.jface.text
 Import-Package: javax.inject;version="1.0.0",
  org.apache.log4j;version="1.2.15",
  org.junit;version="4.5.0",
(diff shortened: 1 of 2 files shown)


LorenzoBettini:
@gallandarakhneorg I checked out this commit into my repository, but there are a few problems:

- [ ] First of all, since this is a test that deals with the content assist, it should not be in the project xtend.core.tests: it should be in the xtend.ide.tests, where you find also a package for tests for the content assist (indeed, also the MANIFEST of xtend.core.tests should not have a dep on jface)
- [ ] related to the above point, I get compilation errors of the shape 
```
Multiple markers at this line
	- ICompletionProposal cannot be resolved to a type
	- The type org.eclipse.jface.text.contentassist.ICompletionProposal cannot be resolved. It is 
	 indirectly referenced from required .class files
```
I think they'll go away once you placed the test in the correct project
- [ ] once you moved the test in the xtend.ide.tests project, subpackage "contentassist" you may have a look at other content assist tests that use a base class, e.g.,
```
public class Bug348446Test extends AbstractXtendContentAssistBugTest
```
or other tests you find in that package, if you find a base class that suits your context.
- [ ] I tried to move the test myself in the ide.tests project and run it as a Junit plug-in tests, but the two tests fail with a NPE...

cdietrich:
@LorenzoBettini @gallandarakhneorg i pushed and fixed this to https://github.com/eclipse/xtext-xtend/tree/gallandarakhneorg-master

thus i propose closing this one and opening a new pullreq. @LorenzoBettini can you have a look at my branch

cdietrich:
replaced by https://github.com/eclipse/xtext-xtend/pull/97

PR #26, opened 2016-08-16
[478377] Add null guard condition

kthoms:
If 'name' feature does not exist for the EClass, a null reference is passed into highlightFeature, which causes a NPE when NodeModelUtils is called.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=478377

meysholdt:
👍

PR #27, opened 2016-08-17
[381031] Add test case testUnusedFunction_1

kthoms:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=381031

meysholdt:
see [eclipse/xtext-extras#41#issuecomment-241000320](/eclipse/xtext-extras/pull/41#issuecomment-241000320)

kthoms:
I think it does not harm to have this test here. Since there are also other tests in the same direction, I would tend to add this case here anyway, but add also test cases to xtext-extras.

meysholdt:
agreed.

kthoms:
Is it OK to merge?

svenefftinge:
+1

PR #30, opened 2016-08-18
Replace usage of classes from org.eclipse.xtext.junit4

kthoms:
This PR changes imports from org.eclipse.xtext.junit4 by classes from
- org.eclipse.xtext.testing
- org.eclipse.xtext.xbase.testing
- org.eclipse.xtext.xbase.junit
Dependencies on these bundles are added. 

The PR does not remove the old dependency, since there are some classes used which are not available in the new base framework. Thus, the dependencies on org.eclipse.xtext.junit4 are just reduced but not resolved now.


meysholdt:
👍 looks good, thank you!

Issue #40, opened 2016-09-14
Ignore optional compile problems on *-gen source folders


kthoms:
Follow-up issue for https://github.com/eclipse/xtext-core/issues/5

kthoms:
Merged with https://github.com/eclipse/xtext-xtend/commit/50570c4b73f74f8c116bc085f7bdfdc203153750

kthoms:
Like for the other gradle based projects the `eclipse-project-layout.gradle` should manage automatic setting of the flag in `.classpath` files.

PR #41, opened 2016-09-14
[#40] Ignore optional compile problems on *-gen source folders

kthoms:

### DIFF
--- a/org.eclipse.xtend.core/.classpath
+++ b/org.eclipse.xtend.core/.classpath
@@ -12,11 +12,13 @@
 	</classpathentry>
 	<classpathentry kind="src" path="src-gen">
 		<attributes>
+			<attribute name="ignore_optional_problems" value="true"/>
 			<attribute name="FROM_GRADLE_MODEL" value="true"/>
 		</attributes>
 	</classpathentry>
 	<classpathentry kind="src" path="emf-gen">
 		<attributes>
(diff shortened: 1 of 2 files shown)


stadlerb:
Wouldn't "optional" be a better fit than "ignore_optional_problems"?

http://help.eclipse.org/neon/index.jsp?topic=%2Forg.eclipse.jdt.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fjdt%2Fcore%2FIClasspathAttribute.html

IGNORE_OPTIONAL_PROBLEMS: If the value of this attribute is "true", all optional problems from the source folder described by this classpath entry will not be reported by the compiler.

OPTIONAL:  If the value of this attribute is "true", the classpath entry is optional. If the underlying resource or jar file doesn't exist, no error is reported and the classpath entry is ignored.

kthoms:
No, it is correct that way.  The folders are not optional.

org.eclipse.xtend.core/.classpath
```diff
@@ -27,6 +29,7 @@
 	</classpathentry>
 	<classpathentry kind="src" path="generator-xtend-gen">
```
LorenzoBettini: If I understand correctly this one is generated by Xtend, so it should not be "ignored" right?

kthoms: Right. Changed.

Issue #43, opened 2016-09-29
[usability] Selection in new files created by the wizard


kthoms:
When an Xtend element is created the selection in the new file spans the newline:

<img width="245" alt="screenshot 31" src="https://cloud.githubusercontent.com/assets/265597/18938953/cb98e1a4-85fc-11e6-84ea-eda56199ea8f.png">

If one would start editing now the line break is deleted.

<img width="247" alt="screenshot 32" src="https://cloud.githubusercontent.com/assets/265597/18938970/e9f882b2-85fc-11e6-964d-890af15c18d6.png">

What is actually desired is that the cursor is placed within the brackets without any selection.

cdietrich:
@kthoms i cannot reproduce this with the current impl. do you have any hints how to reproduce?

kthoms:
I'll come back to you tomorrow.

Am 01.11.2016 um 11:35 schrieb Christian Dietrich <notifications@github.com>:

kthoms:
Checked again. Strangely I cannot observe the issue in my developer workspace either. But it becomes reproducible when starting the runtime environment from the developer workspace, and then the patch fixes the issue. Maybe there is something different in the underlying behavior of the base plugins and a difference between developer IDE and target platform.

Please start a runtime workspace and create an Xtend class. Then you should face the strange selection.

cdietrich:
@kthoms i today saw a different behaviour, how the first tab is marked.

<img width="149" alt="bildschirmfoto 2017-01-13 um 16 16 05" src="https://cloud.githubusercontent.com/assets/238736/21934925/a789fb10-d9ab-11e6-9877-f0b5f0ddec87.png">

kthoms:
Reproducible on Oxygen target, not on Luna. I'll try to find a solution that works on both.

kthoms:
Hmm, to be more precise, it works for me also when setting Oxygen as target. But when I'm Oomphing an Oxygen distro and install Xtend on it, I can see your issue.
The Oxygen target has slightly more recent version of JFace, but I'm not sure if there is anything in there that fixes this.
So in my devenv I can't reproduce it yet, I'll observe that further.

cdietrich:
fixed in Xtext 2.13

PR #44, opened 2016-09-29
[#43] Set cursor without selection after opening the editor for a new file

kthoms:
With this change the cursor is placed in the element body without selection:

screenshot 33


spoenemann:
Why is it different from the current behavior? If I create an Xtend class with v. 2.10.0 I get exactly what you described.

cdietrich:
@spoenemann @kthoms maybe this is a luna vs neon thing

kthoms:
I still have that issue. This is how a new Xtend class looks like. If you start editing then the line break in the class body would disappear. Target platform is Luna.
screenshot 140

kthoms:
With the fix the issue disappears.

screenshot 141

kthoms:
Rebased the branch on master.

cdietrich:
@kthoms can you give step by step what you do?

kthoms:
Just create a new Xtend class with the New Xtend Class wizard

cdietrich:
i cannot reproduce this i a runtime app nor a runtime app started from a runtime app

kthoms:
I took both screenshots today, and you see the different behavior. So I propose that if my fix does not harm your case, then it seems a valid change. I can reproduce the issue and also the fix. No idea what is different here, but I'm using an Oomph provisioned devenv with Luna target.

spoenemann:
I'd say let's merge this and test the issue on different platforms with the resulting update site.

PR #45, opened 2016-09-29
[378821] Add option to generate main method

kthoms:
Option added like in JDT's class wizard:

screenshot 34

Created file:
screenshot 35


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=378821

org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/wizards/XtendTypeCreatorUtil.java
```diff
@@ -90,6 +91,15 @@ public static String createClassContent(String typeName, String superClass,
 		sb.append(" {");
 		sb.append(lineSeparator);
 		sb.append(indentation);
+		if (createMain) {
+			sb.append("def static void main (String[] args) {");
```
cdietrich: is the space between main and () intended? the formatter does not add it

kthoms:
You are right, the formatter would not add the space. So the wizard shouldn't do that also.
Changed.

cdietrich:
thanks

PR #47, opened 2016-09-30
Kth/xtend gen

kthoms:
Changes due to use of the latest Xtend compiler 2.11.0.v20160929-1448


kthoms:
Partially obsolete by https://github.com/eclipse/xtext-xtend/pull/80
New changes will be done in separate PR.

Issue #46, opened 2016-09-29
consider to use mwe2 0.0.0 in target file


cdietrich:
the xtend target uses a specific mwe2 version which leads to build failures if the version timestamp changes once a month

so we should either have a update site that conserves the old versions for mwe2 or use 0.0.0 as version as we do it for the other stuff

PR #51, opened 2016-10-11
always use latest greatest mwe2 version. solves #46

cdietrich:
always use latest greatest mwe2 version. solves #46


cdietrich:
please review

meysholdt:
thx!

PR #61, opened 2016-10-21
Typo in class name fixed

kthoms:


cdietrich:
@meysholdt is this change ok with respect to API? what do you think?

svenefftinge:
+1

PR #66, opened 2016-10-31
[#6] fixed some tests after java 8 migration

cdietrich:
[#6] fixed some tests after java 8 migration


cdietrich:
please review. i compared the code in java 7 and java 8 and saw that java 7 uses api that is now ambigous in java 8, thus i changed the calls and expections to avoid the conflict. thus the test should be still testing the same "thing"

PR #82, opened 2016-12-05
[eclipse/xtext#1089] Replace (not)equal operators by triple (not)equals

kthoms:
For comparison with null the triple (not) equal operator should be used. 
Resolved compiler warnings
Regenerated xtend-gen


kthoms:
I have been more careful with the changes. For review please check that no place was changed where the operators are in template expressions. In the generated Java code this can be seen if a change in a line with buffer.append() calls.

kthoms:
Thx!

Issue #92, opened 2016-12-09
Fix xtend-archetype integration test


spoenemann:
The changes done in #93 caused the xtend-archetype integration test to fail. It's probably related to the fact that xtend.core is no longer built in the same build.

The integration test is currently disabled with the option `-Dit-tests-skip=true` in the Jenkinsfile.

On the other hand, #18 seems to work again.

cdietrich:
@spoenemann do both builds still use the same local maven repo?

cdietrich:
it seems you did not adapt code like this one String localRepo = new File("../local-maven-repository/").getAbsolutePath();

cdietrich:
additionally the gradle stuff is put to different repo

PR #94, opened 2016-12-09
Fixed Xtend*IT. Solves #92

cdietrich:


cdietrich:
@spoenemann what do you think?

spoenemann:
Looks good, thanks for the help!

Issue #29, opened 2016-08-18
[tests] Replace usage of org.eclipse.xtext.junit4 by org.eclipse.xtext.testing


kthoms:
The Xtend tests are still using the deprecated org.eclipse.xtext.junit4 bundle, they should use the org.eclipse.xtext.testing bundle now.

kthoms:
<h1>Replacements</h1>
<table>
<tr><th>old class</th><th>new class</th></tr>
<tr><td>org.eclipse.xtext.junit4.InjectWith</td><td>org.eclipse.xtext.testing.InjectWith</td></tr>
<tr><td>org.eclipse.xtext.junit4.XtextRunner</td><td>org.eclipse.xtext.testing.XtextRunner</td></tr>
<tr><td>org.eclipse.xtext.junit4.logging.LoggingTester</td><td>org.eclipse.xtext.testing.logging.LoggingTester</td></tr>
<tr><td>org.eclipse.xtext.junit4.util.ResourceHelper</td><td>org.eclipse.xtext.testing.util.ResourceHelper</td></tr>
<tr><td>org.eclipse.xtext.junit4.validation.ValidationTestHelper</td><td>org.eclipse.xtext.testing.validation.ValidationTestHelper</td></tr>
<tr><td>org.eclipse.xtext.junit4.internal.LineDelimiters</td><td>org.eclipse.xtext.xbase.junit.LineDelimiters</td></tr>
<tr><td>org.eclipse.xtext.junit4.TemporaryFolder</td><td>org.eclipse.xtext.xbase.testing.TemporaryFolder</td></tr>
<tr><td>org.eclipse.xtext.junit4.formatter.FormatterTester</td><td>org.eclipse.xtext.testing.formatter.FormatterTestHelper</td></tr>

</table>

<h1>Missing framework classes</h1>
Some classes used from xtext.junit4 are not available in xtext.testing. I assume they should be copied.
- org.eclipse.xtext.junit4.util.InMemoryURIConverter
- org.eclipse.xtext.junit4.internal.StopwatchRule
- org.eclipse.xtext.junit4.internal.Timed

Until these classes are available, the tests have to rely on both bundles.

org.eclipse.xtend.core.tests.typesystem.ErrorTest.ResolveBatchedWithoutJRE extends org.eclipse.xtext.xbase.junit.typesystem.NoJRESmokeTester, which itself depends on xtext.junit4. This has also to be refactored in a separate issue.

meysholdt:
[...]

PR #96, opened 2016-12-13
get rid of xtext.junit4. solves #29

cdietrich:
get rid of xtext.junit4. solves #29


cdietrich:
@spoenemann can you please have a look

PR #97, opened 2016-12-14
Adding tests for dirty state type modifiers.

cdietrich:
see eclipse/xtext-eclipse#28
see eclipse/xtext-eclipse#30

replaces https://github.com/eclipse/xtext-xtend/pull/23

### DIFF
--- a/org.eclipse.xtend.core.tests/META-INF/MANIFEST.MF
+++ b/org.eclipse.xtend.core.tests/META-INF/MANIFEST.MF
@@ -21,8 +21,7 @@ Require-Bundle: org.eclipse.xtend.core,
  org.objectweb.asm;bundle-version="5.0.1",
  org.eclipse.jdt.core,
  org.eclipse.xtext.builder.standalone,
- org.eclipse.xtext.java,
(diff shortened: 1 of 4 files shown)


cdietrich:
please review

LorenzoBettini:
@cdietrich If I understand correctly, this PR does not only add a test but also introduces the fix? Or must the actual fix be taken directly from https://github.com/eclipse/xtext-eclipse/pull/30 ?

LorenzoBettini:
@cdietrich I give the answer myself :) your tests succeed once I also take the commit of https://github.com/eclipse/xtext-eclipse/pull/30 (while one of the new tests fail without that). I'm running a Jenkins build for https://github.com/eclipse/xtext-eclipse/pull/30 just to make sure it does not break anything else, then I think we could merge https://github.com/eclipse/xtext-eclipse/pull/30 and then this PR, is that OK?

cdietrich:
ok ill have a look why the ip validaiton complains

LorenzoBettini:
The build works so I'll merge

PR #100, opened 2016-12-19
Applied fixes from branch milestone_2.11.0.beta2

spoenemann:


spoenemann:
Current test failures seem unrelated to the Maven plugin configuration.

PR #111, opened 2017-01-04
Compile sources from folders longrunning and smoke-suites

kthoms:
The folders
- longrunning/src
- longrunning/xtend-gen
- smoke-suites
were not configured as source folders properly. When imported as Gradle
project this requires that the dependencies on Equinox and PDE can be
resolved by Gradle.

Note: The new Buildship version seems to change the format of .classpath files. The file change was caused by a project re-import.


cdietrich:
I thought all the time the previous strange config was intended by @spoenemann

kthoms:
Maybe, but I am not aware of a reason.

cdietrich:
Maybe cause the tests were executed by gradle as well before

kthoms:
Hmm, maybe. But the Gradle build seems not to execute them, see http://services.typefox.io/open-source/jenkins/job/xtext-xtend/job/kth_core_tests_missing_sourcesets/

spoenemann:
We should either make xtend.idea.tests independent of xtend.core.tests, which would allow to remove xtend.core.tests from the Gradle build so it's a pure PDE project, or extract the tests with PDE dependencies, e.g. move them to xtend.ui.tests.

cdietrich:
@kthoms needs a rebase

spoenemann:
Now that the Eclipse bundles are properly published on Maven Central, we should use the new group ids and versions, e.g. `org.eclipse.pde:org.eclipse.pde.core:3.11.1`

@kthoms could you update accordingly?

spoenemann:
Sorry, I forgot that now with the Xtend compiler running in the Gradle build, the configuration of the additional source folders is more complex due to the different Xtend compiler outputs. I copied the configuration from xbase.tests. Now all source folders are included and the bundle is free of compile errors in my workspace.

PR #117, opened 2017-01-10
Align the version of jdt.core in maven to the one from gradle

holgerschill:
Signed-off-by: Holger Schill <Holger.Schill@itemis.de>

### DIFF
--- a/releng/org.eclipse.xtend.tycho.parent/pom.xml
+++ b/releng/org.eclipse.xtend.tycho.parent/pom.xml
@@ -24,8 +24,8 @@
 		<log4j.version>1.2.16</log4j.version>
 		<commons-logging.version>1.1.3</commons-logging.version>
 		<commons-cli.version>1.2</commons-cli.version>
-		<equinox-common-version>3.6.0</equinox-common-version>
-		<jdt-core-version>3.10.0</jdt-core-version>
+		<equinox-common-version>3.8.0</equinox-common-version>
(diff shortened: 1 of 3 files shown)


releng/org.eclipse.xtend.tycho.parent/pom.xml
```diff
@@ -24,8 +24,8 @@
 		<log4j.version>1.2.16</log4j.version>
 		<commons-logging.version>1.1.3</commons-logging.version>
 		<commons-cli.version>1.2</commons-cli.version>
-		<equinox-common-version>3.6.0</equinox-common-version>
-		<jdt-core-version>3.10.0</jdt-core-version>
+		<equinox-common-version>3.8.0</equinox-common-version>
+		<jdt-core-version>3.12.2</jdt-core-version>
```
spoenemann: See my comment on eclipse/xtext-eclipse#152

holgerschill: Even when the properties are not used we could merge that, right?

spoenemann: Yes that wouldn't break the build, but it would be better to remove unused properties.

holgerschill: I'll remove them later. Would be good to have the gradle stuff in the repo.

spoenemann: 👍

PR #121, opened 2017-01-11
[batch compiler] add compiler.apt dependency (fixes #116)

### LINKED ISSUE #116: Xtend-Maven Plugin does no longer work
Xtend-Maven Plugin does no longer work
see
https://github.com/cdietrich/xtext-211-maven
for an example

i assume there are some changes to the maven metadata or java sourcelang that cause this

svenefftinge:


cdietrich:
It's actually extras when consuming from maven one might run in the same problem

svenefftinge:
It seems to be only needed when using the XtendBatchCompiler. At least I don't have issues with the gradle plugin and the java source language.

cdietrich:
No I mean when consuming from maven.
The problem with gradle and maven is how gradle generates transitive optional dependencies to the generated poms. (I don't know what's exactly in the jdt maven metadata)

PR #128, opened 2017-01-16
[460455] Add test cases

kthoms:
This adds tests for the erroneus cases described in https://bugs.eclipse.org/bugs/show_bug.cgi?id=460455

The first case is addressed by https://github.com/eclipse/xtext-extras/pull/105.
The second case is not fixed yet and therefore disabled. Likely it has to be solved in ConflictResolver.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=450655

nyssen:
I think the issue tracker reference is wrong (it points to a GEF bug).

kthoms:
Yes, I already fixed the commit.

PR #147, opened 2017-02-01
fixed extract method refactoring. 

cdietrich:
fixed extract method refactoring. 

resolves https://github.com/eclipse/xtext-extras/issues/107

### DIFF
--- a/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/refactoring/ExtractMethodRefactoring.java
+++ b/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/refactoring/ExtractMethodRefactoring.java
@@ -331,6 +331,10 @@ public class ExtractMethodRefactoring extends Refactoring {
 		try {
 			Set<String> calledExternalFeatureNames = newHashSet();
 			returnType = calculateReturnType(resolvedTypes);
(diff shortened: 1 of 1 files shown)


cdietrich:
please review

kthoms:
This would fix the NPE, but the actual problem is that the return type is not computed correctly. Consider this file:
```
package test

import java.util.ArrayList
import org.eclipse.xtend.lib.annotations.Data

class Foo {
	Bar thisEnd
	def boolean test() {
		val int a = 5
		val x = '''«thisEnd.name»'''
		return new ArrayList<Integer>.addAll(#[a])
	}
}

@Data
class Bar {
	String name
}
```
If you extract the two lines `val x = ...` and `return ...`, then the refactoring would now determine void as the type. This will lead to a compile error, since the extracted method has a return statement with a type expectation, but it is incompatible with the declared return type void.

cdietrich:
can we moved this discussion to the ticket. i hope @JanKoehnlein or @svenefftinge find the time to look at. since i dont know how to common supertype calculation is intended to work

svenefftinge:
I will come up with a better fix

svenefftinge:
See : https://github.com/eclipse/xtext-xtend/pull/190

PR #161, opened 2017-02-23
[457930] Better recording of non resolved imported names in xtend

cdietrich:
[457930] Better recording of non resolved imported names in xtend

### DIFF
--- a/org.eclipse.xtend.core/src/org/eclipse/xtend/core/scoping/RecordingTypeScope.java
+++ b/org.eclipse.xtend.core/src/org/eclipse/xtend/core/scoping/RecordingTypeScope.java
@@ -10,7 +10,9 @@ package org.eclipse.xtend.core.scoping;
 import java.util.Set;
 
 import org.eclipse.emf.ecore.EObject;
+import org.eclipse.xtext.common.types.access.impl.AbstractJvmTypeProvider.ClassNameVariants;
 import org.eclipse.xtext.common.types.xtext.AbstractTypeScope;
+import org.eclipse.xtext.naming.IQualifiedNameConverter;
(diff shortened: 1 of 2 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=457930

org.eclipse.xtend.core/src/org/eclipse/xtend/core/scoping/RecordingTypeScope.java
```diff
@@ -23,22 +25,38 @@
 	private final AbstractTypeScope typeScope;
 	private final Set<QualifiedName> importedNames;
 
-	public RecordingTypeScope(AbstractTypeScope typeScope, Set<QualifiedName> importedNames) {
-		super(null, null, null);
+	public RecordingTypeScope(AbstractTypeScope typeScope, Set<QualifiedName> importedNames, IQualifiedNameConverter qnc) {
```
kthoms: This will break API.
You have to create a second constructor, mark it with @since 2.12 and make the original deprecated.

cdietrich: i dont think a package private class is API is it?

holgerschill: For me that class is not API as it is not public.

kthoms: Right, it isn't public. So ok.

PR #107, opened 2016-12-24
[492072] Adjust test expectations

kthoms:
This will change the test expectations for the Xtend compiler changed by eclipse/xtext-extras#91.

With eclipse/xtext-extras#91 many unnecessary variables disappear in the compiled Xtend code.

PR #190, opened 2017-03-29
[refactoring] Fix problems with return type in extract method (fixes #107)

svenefftinge:


svenefftinge:
The underlying problem, was that IResolvedTypes#getReturnType(XExpression) returns the actual type if no return type exists.

Issue #191, opened 2017-03-29
Invalid support in implementation of the default methods.


gallandarakhneorg:
Let the following Xtend code:
```
class TestMultiInheritance<T> extends AbstractCollection<T> implements Set<T> {
	
	override iterator() {
		spliterator
                null
	}
	
	override size() {
               0
	}
	
}
```

The Xtend compiler generates the error "The type TestMultiInheritance inherits multiple implementations of the method spliterator() from Iterable and Set."

However, the equivalent Java code is compiling:
```java
public class TestMultiInheritance<T extends Object> extends AbstractCollection<T> implements Set<T> {
  public Iterator<T> iterator() {
    this.spliterator();
    return null;
  }
  
  public int size() {
    return 0;
  }
}
```

The spliteraor function is defined in the Collection and Set interfaces with different code. The Java compiler takes the default implementation from Set because is it the closest definition from the call in the type hierarchy.

Is a feature of Xtend to fail in this case? Or should Xtend be fixed in order to have a similar behavior as the Java compiler?

cdietrich:
looks like a bug in the return type inference to me.

```
class TestMultiInheritance<T> extends AbstractCollection<T> implements Set<T> {
	
	override Iterator<T> iterator() {
		spliterator
                null
	}
	
	override size() {
               0
	}
	
}
```
seems to work

cdietrich:
wthat makes me wonder: which super method wins in java?

gallandarakhneorg:
The one coming from the just implemented interface. As far as I remember, the heuristic is based on the distance between the call and the default method definition into the type hierarchy. The definition of spliterator in Set is closer to the definition of spliterator in Collection or Iterable (provided by the extended type).

I just created the example in my Eclipse for DSL developer (not my own DSL product), and the issue is here.

cdietrich:
yes unfortunately we have tons of bugs around handling of default methods

gallandarakhneorg:
I will search for a fix and create a PR if I found something suitable.

PR #192, opened 2017-03-29
Ignore default methods when checking ambiguous inherited methods.

gallandarakhneorg:
see eclipse/xtext#2296


org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/ValidationIssue191Test.xtend
```diff
@@ -0,0 +1,44 @@
...
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtend.core.tests.validation
+
+import javax.inject.Inject
+import org.eclipse.xtend.core.tests.AbstractXtendTestCase
+import org.eclipse.xtend.core.xtend.XtendFile
+import org.eclipse.xtext.testing.util.ParseHelper
+import org.eclipse.xtext.testing.validation.ValidationTestHelper
+import org.eclipse.xtext.xbase.XbasePackage
```
cdietrich: unused import

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/ValidationIssue191Test.xtend
```diff
@@ -0,0 +1,44 @@
...
+import org.eclipse.xtext.xbase.validation.IssueCodes
```
cdietrich: unused import

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/ValidationIssue191Test.xtend
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.Test
+
+/**
+ * @author Stephane Galland - Initial contribution and API
+ */
+class ValidationIssue191Test extends AbstractXtendTestCase {
```
cdietrich: Test fails locally: 

```
java.lang.AssertionError: Expected no errors, but got :ERROR (org.eclipse.xtext.diagnostics.Diagnostic.Syntax) 'mismatched input ':' expecting '}'' on XtendClass, offset 137, length 1
ERROR (org.eclipse.xtend.core.validation.IssueCodes.class_must_be_defined_abstract) 'The class XXX must be defined abstract because it does not implement size()' on XtendClass, offset 63, length 3
ERROR (org.eclipse.xtend.core.validation.IssueCodes.missing_override) 'The method iterator() of type XXX must use override keyword since it actually overrides a supertype method.' on XtendFunction, offset 126, length 8
ERROR (org.eclipse.xtend.core.validation.IssueCodes.missing_abstract) 'The abstract method iterator in type XXX can only be defined by an abstract class.' on XtendFunction, offset 126, length 8
ERROR (org.eclipse.xtend.core.validation.IssueCodes.abstract_method_missing_return_type) 'The abstract method iterator in type XXX must declare a return type' on XtendFunction, offset 126, length 8
```

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/ValidationIssue191Test.xtend
```diff
@@ -0,0 +1,44 @@
...
+	@Inject extension ValidationTestHelper
+	@Inject extension ParseHelper<XtendFile>
+
+	@Test	
+	def void testIssue191() {
+		val file = '''
+			import java.util.Set
+			import java.util.AbstractCollection
+			class XXX<T> extends AbstractCollection<T> implements Set<T> {
+			  def iterator() : Iterator<T> {
```
cdietrich: i assume you mean something like

```
class XXX<T> extends AbstractCollection<T> implements Set<T> {
	override Iterator<T> iterator() {
		null
	}

	override int size() {
		0
	}
}

```

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/ValidationIssue191Test.xtend
gallandarakhneorg: The syntax of the tested Xtend code was fixed.

cdietrich: there is still a big bunch of failing tests
http://services.typefox.io/open-source/jenkins/job/xtext-xtend/job/gallandarakhneorg_master/1/testReport/

cdietrich:
still failing tests http://services.typefox.io/open-source/jenkins/job/xtext-xtend/job/gallandarakhneorg_master/2/#showFailuresLink

gallandarakhneorg:
I have fixed all the failing unit tests.
Please check if my changes in the `Java8ValidationTest.xtend` file fit your expectations.

szarnekow:
To be honest, I wouldn't expect any changes to existing test cases. Please double check that your fixes match the Java language spec for default methods. At least the existing test cases should pretty much reflect that one.

gallandarakhneorg:
According to my understanding of JLS (https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html), the Java8ValidationTest were not respecting all the specification related to inheritance: errors were generated in Xtend when Java compiler does not.
This PR should provides a fix to the detected issue.

org.eclipse.xtend.core.tests.java8/src/org/eclipse/xtend/core/tests/java8/validation/Java8ValidationTest.xtend
```diff
@@ -76,8 +77,7 @@ class Java8ValidationTest extends AbstractXtendTestCase {
 				def void foo() { }
 			}
 			class C implements A, B { }
-		''').assertError(XTEND_CLASS, CONFLICTING_DEFAULT_METHODS,
-			"The type C inherits multiple implementations of the method foo() from A and B.")
+		''').assertNoErrors
```
szarnekow: We inherit two default methods. Why is that not an error anymore?

org.eclipse.xtend.core.tests.java8/src/org/eclipse/xtend/core/tests/java8/validation/Java8ValidationTest.xtend
```diff
@@ -90,8 +90,7 @@ class Java8ValidationTest extends AbstractXtendTestCase {
 				def void foo() { }
 			}
 			interface C extends A, B { }
-		''').assertError(XTEND_INTERFACE, CONFLICTING_DEFAULT_METHODS,
-			"The type C inherits multiple implementations of the method foo() from A and B.")
+		''').assertNoErrors
```
szarnekow: Same here.

org.eclipse.xtend.core.tests.java8/src/org/eclipse/xtend/core/tests/java8/validation/Java8ValidationTest.xtend
szarnekow: This looks wrong to me. According to the spec:
It is a compile-time error if a class C inherits a default method whose signature is override-equivalent with another method inherited by C, unless there exists an abstract method declared in a superclass of C and inherited by C that is override-equivalent with the two methods.

org.eclipse.xtend.core.tests.java8/src/org/eclipse/xtend/core/tests/java8/validation/Java8ValidationTest.xtend
szarnekow: Same here:
If an interface I inherits a default method whose signature is override-equivalent with another method inherited by I, then a compile-time error occurs. (This is the case whether the other method is abstract or default.)

szarnekow (commented):
It's certainly possible that there are gaps in the impl, but the changed test expectations look certainly bogus to. I just put two examples in the review comments.

szarnekow:
Do you still plan to work on this PR?

gallandarakhneorg:
Yes. But due to my stress schedule, I'm able to work one this PR to obtain something at the end of September only.

gallandarakhneorg:
According to my experiments, I found two test cases within `Java8ValidationTest` which have different results in Xtend and Java.

* In `testConflictingDefaultMethods05` function, the Java equivalent code could be compiled:
```java
public class Test {
	interface A {
		default int foo(java.util.List<String> list) { return 0; }
	}
	interface B {
		default double foo(java.util.List<Class<?>> list) { return 0; }
	}
	class C implements A, B { }
	private void t(C c) {
		java.util.List<String> list1 = null;
		java.util.List<Class<?>> list2 = null;
		c.foo(list1);
		c.foo(list2);
	}
}
```

* In `testConflictingDefaultMethods06` function, the Java equivalent code could be compiled:
```java
public class Test {
	interface A {
		default int foo(java.util.List<String> list) { return 0; }
	}
	interface B {
		default double foo(java.util.List<Class<?>> list) { return 0; }
	}
	interface C extends A, B { }
	private void t(C c) {
		java.util.List<String> list1 = null;
		java.util.List<Class<?>> list2 = null;
		c.foo(list1);
		c.foo(list2);
	}
}
``` 

Additionally, the case which is at the source of my patch does not compile with Xtend, but with Java it is ok:
```java
private class ResolvedByAbstractType01 {
	class A<T> extends java.util.AbstractCollection<T> implements java.util.Set<T> {
		public java.util.Iterator<T> iterator() {
			return null;
		}
		public int size() {
			return 0;
		}
	}
}
```

I you are agree with my results, I could provide a patch of the XtendValidator.

gallandarakhneorg:
The question is: what is the best class to update?:
* `XtendValidator.doCheckOverriddenMethods`,
* `ResolvedFeatures.getAllOperations`.

szarnekow:
That would be `ResolvedFeatures`. Please add links to the relevant part of the Java spec to your commit message if you make changes there.

szarnekow:
Please also check against javac, because your first example appears to be rejected.

`Error:(41, 3) java: name clash: foo(java.util.List<java.lang.Class<?>>) in Test.B and foo(java.util.List<java.lang.String>) in Test.A have the same erasure, yet neither overrides the other`

gallandarakhneorg:
I agree that  all the existing tests are correct, against the use of `javac` (I missed to test with it). Do not consider my comments related to them.

Third case is valid with `javac` and generates an error with Xtend.

gallandarakhneorg:
Section [8.4.8](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.8) of JLS clearly states that a class inherits the body of the default methods from its direct superinterfaces.

In the third case, the detected conflict is on `splititerator` between `Collection` and `Set`. But `Collection` is not a direct superinterface of `A`. `Set` should wins and provides its default method implementation.

gallandarakhneorg:
If you are agree with my interpretation of JLS, I could work on a patch.

gallandarakhneorg:
Changes of `ResolvedFeatures` class is included within the [Xbase PR 193](https://github.com/eclipse/xtext-extras/pull/193).

gallandarakhneorg:
According to discussions within https://github.com/eclipse/xtext-extras/pull/193, I will create 8748 unit tests for validating that the Xtend compiler follows the Java specification regarding the inheritance and implementation features. Each unit test is based on a small Java code that is compiled successfully or not by javac. This success flag is used as the expected output of the Xtend compiler.

gallandarakhneorg:
My first experiments indicates that 760 unit tests over the 8748 are failing due to 'Duplicate implemented feature' of type Interface.

gallandarakhneorg:
Should we considerer that the error generated by the Xtend compiler is still acceptable, even not generated by the Java compiler?
If yes, the patch provided by [xtext-extras PR 193](https://github.com/eclipse/xtext-extras/pull/193/files) fixes the detected issues in #191.
If not, I propose to change the criticity level from error to (ignorable) warning.

gallandarakhneorg:
Currently, I found 60 cases other the 8748 in which the behaviors of the javac and the Xtend compilers are not compatible.

Below I show you three examples in which the Xtend compiler and javac are not agree.

First example's Java code:
```java
public interface I2 {
  public void fct(int arg);
}
public interface I1 {
  public default void fct(int arg) {}
}
public abstract static class IC implements I2 {
}
public interface I3 extends I1 {
  public void fct(int arg);
}
public static class C extends IC implements I1, I3 {
}
```
In the previous code, the I3 interface hides the default implementation from I1. As expected javac provides "C is not abstract and does not override abstract method fct(int) in I2". But Xtend provides no error message.

Second example's Java code:
```java
public interface I2 {
  public void fct(int arg);
}
public interface I1 extends I2 {
}
public abstract static class IC implements I1 {
}
public interface I3 extends I2 {
  public default void fct(int arg) {}
}
public static class C extends IC implements I1, I3 {
}
```
With the previous code, javac gives no error message. But, Xtend gives "The non-abstract method fct(int) inherited from I3 conflicts with the method fct(int) inherited from I2".

Third example's Java code:
```java
public interface I2 {
  public default void fct(int arg) {}
}
public interface I1 extends I2 {
}
public abstract static class IC {
}
public interface I3 extends I1 {
  public default void fct(int arg) {}
}
public static class C extends IC implements I1, I3 {
}
```
With the previous code, javac gives no error message. But, Xtend gives "The type C inherits multiple implementations of the method fct(int) from I2 and I3".

szarnekow:
Nice! Are you working on the fixes for the remaining tests?

gallandarakhneorg:
First, I will push to this PR the unit tests in order to let all to discuss about its coverage (that is not perfect for sure). Then, I will work on the patch (do not expect a patch before end of February).

szarnekow:
I'm afraid that was a misunderstanding. It's not that the coverage is not sufficient, but my impression was that the fixes do not make all the tests green yet, do they?

gallandarakhneorg:
Dear all.
I'm still interested by this PR.
The current test coverage is supposed to highlight differences between the inheritance mechanisms in Java and the ones in Xtend. If a test is failing (and assuming I did not code it in a bad way), it means that the Java compiler and the Xtend compiler provide a different answer regarding the compilation of similar codes.

szarnekow:
I'll take this one for one of the next releases

cdietrich:
move to 2.24

cdietrich:
this repository has been merged/moved to https://github.com/eclipse/xtext.
please feel free to recreate this PR against that repository in case it still is relevant.

PR #193, opened 2017-03-29
Enable variable access inside lambda.

gallandarakhneorg:
Variable accesses are allowed when the generator is using at least Java8.


svenefftinge:
Xtend doesn't have the effectively final feature of Java 8, so this is not valid.

PR #199, opened 2017-04-07
Sort References on Rename Refactoring to Prevent NPE

cdietrich:
Sort References on Rename Refactoring to Prevent NPE when Anon Inner Class Constructors are Involved. Fixes https://github.com/eclipse/xtext-core/issues/300


JanKoehnlein:
Does Xbase support local classes with an `@localclasses` URI at all? Apart from that, it looks good.

cdietrich:
afaik xbase does not support it. this is why i put it here

JanKoehnlein:
My bad. I missread the class names.

Issue #206, opened 2017-04-28
Annotations on annotation properties are not working


Sandared:
```
@ObjectClassDefinition
annotation MyConfiguration{

    @AttributeDefinition
    String someProp = 'test'

}
```

Xtend complains about "The annotation @AttributeDefinition is disallowed for this location", but if I modify the generated Java file there is no error shown:

```
@ObjectClassDefinition
public @interface MyConfiguration {

    @AttributeDefinition
    public String someProp() default "test";

}
```

cdietrich:
annotation looks like
```
import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;

@Retention(value = RetentionPolicy.CLASS)
@Target(value = ElementType.METHOD)
public @interface AttributeDefinition {
}

```

xtend validator does

```
final EObject eContainer = getContainingAnnotationTarget(annotation);
		for (Entry<Class<?>, Collection<ElementType>> mapping : targetInfos.asMap().entrySet()) {
```

eContainer (=the annotated object) is a XtendField.

the map is populated as:

```
result.put(XtendField.class, ElementType.FIELD);
```

`ElementType.FIELD` != `ElementType.METHOD`

=> i guess we need some special treatment for xtendfields in annotations

cdietrich:
fixed in xtext 2.12

PR #208, opened 2017-05-04
 fixed broken annotation target type validation with fields in xtend-annotations #206

cdietrich:
fixed broken annotation target type validation with fields in xtend-annotations #206


cdietrich:
please review

PR #209, opened 2017-05-05
[409297][hightlighting][deprecation] Semantics different from JDTs

cdietrich:
[409297][hightlighting][deprecation] Semantics different from JDTs


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=409297

Issue #218, opened 2017-05-30
Content Assist after fields with Annotations is Broken


cdietrich:
Here Content Assist does work

```
class OK {
	MLCDP| // resolves to MultiLineCommentDocumentationProvider
}
```

here it does not

```
class Bad {
	@Inject MLCDP| // rno proposal
}
```

cdietrich:
in the log/console

```
288172 [main] WARN  org.eclipse.xtext.ui.editor.contentassist.AbstractJavaBasedContentProposalProvider  - Error in polymorphic dispatcher : org.eclipse.xtext.impl.RuleCallImpl cannot be cast to org.eclipse.xtext.Keyword
java.lang.ClassCastException: org.eclipse.xtext.impl.RuleCallImpl cannot be cast to org.eclipse.xtext.Keyword
	at org.eclipse.xtend.ide.contentassist.XtendProposalProvider.isExtendsClause(XtendProposalProvider.java:245)
	at org.eclipse.xtend.ide.contentassist.XtendProposalProvider.completeJvmParameterizedTypeReference_Type(XtendProposalProvider.java:235)
	at sun.reflect.GeneratedMethodAccessor127.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.eclipse.xtext.util.PolymorphicDispatcher.invoke(PolymorphicDispatcher.java:302)
	at org.eclipse.xtext.ui.editor.contentassist.AbstractJavaBasedContentProposalProvider.invokeMethod(AbstractJavaBasedContentProposalProvider.java:300)
	at org.eclipse.xtext.ui.editor.contentassist.AbstractJavaBasedContentProposalProvider.completeAssignment(AbstractJavaBasedContentProposalProvider.java:205)
	at org.eclipse.xtext.ui.editor.contentassist.AbstractContentProposalProvider$DefaultContentAssistProcessorSwitch.caseAssignment(AbstractContentProposalProvider.java:66)
	at org.eclipse.xtext.ui.editor.contentassist.AbstractContentProposalProvider$DefaultContentAssistProcessorSwitch.caseAssignment(AbstractContentProposalProvider.java:1)
	at org.eclipse.xtext.util.XtextSwitch.doSwitch(XtextSwitch.java:144)
[...]

PR #219, opened 2017-05-30
fixed ClassCastException preventing Type Proposals in fields from working

cdietrich:
fixed ClassCastException preventing Type Proposals in fields from working #218


cdietrich:
please review

PR #224, opened 2017-06-13
fixed compile errors through transitive optionals

cdietrich:


cdietrich:
otherwise i get compile errors on ASTParserFactory

Java problem: The type org.eclipse.core.runtime.Plugin cannot be resolved. It is indirectly referenced from required .class files

cdietrich:
please review

PR #235, opened 2017-07-26
[467063] allow creating default methods from active annotations

cdietrich:
[467063] allow creating default methods from active annotations

### DIFF
--- a/org.eclipse.xtend.core/src/org/eclipse/xtend/core/macro/declaration/JvmBasedDeclarations.xtend
+++ b/org.eclipse.xtend.core/src/org/eclipse/xtend/core/macro/declaration/JvmBasedDeclarations.xtend
@@ -1028,6 +1028,9 @@ class MutableJvmMethodDeclarationImpl extends JvmMethodDeclarationImpl implement
 	override setAbstract(boolean isAbstract) {
 		checkMutable
 		delegate.setAbstract(isAbstract)
+		if (isAbstract) {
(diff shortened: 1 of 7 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=467063

org.eclipse.xtend.core/src/org/eclipse/xtend/core/macro/declaration/JvmBasedDeclarations.xtend
```diff
@@ -1052,6 +1052,10 @@ class MutableJvmMethodDeclarationImpl extends JvmMethodDeclarationImpl implement
 	override setDefault(boolean isDefault) {
 		checkMutable
 		delegate.setDefault(isDefault)
+		if (isDefault) {
+			delegate.abstract = false
```
szarnekow: Shoudl we also set default to false when we set abstract to true?

szarnekow:
Looks good.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/macro/declaration/JvmBasedDeclarations.xtend
cdietrich: adapted (same for static)

PR #250, opened 2017-09-01
[434688, 505760] Fixed 'Show In/ Generated Code'

kthoms:
Provide adapter IShowInTargetList for the editor through adapter factory
Implemented IShowInTarget for DerivedSourceView
Removed bogus menu contribution (parameters are not supported)

### DIFF
--- a/org.eclipse.xtend.ide/plugin.xml
+++ b/org.eclipse.xtend.ide/plugin.xml
@@ -229,18 +229,6 @@
       </command>
     </extension>
     <extension point="org.eclipse.ui.menus">
-         <menuContribution locationURI="popup:org.eclipse.ui.menus.showInMenu">
-            <command commandId="org.eclipse.ui.views.showView" style="push"
-                    icon="icons/generated_code_view.png" label="%DerivedSourceView">
-                <parameter name="org.eclipse.ui.views.showView.viewId"
-                        value="org.eclipse.xtend.ide.view.DerivedSourceView">
-                </parameter>
-                <visibleWhen checkEnabled="false">
-                        <reference definitionId="org.eclipse.xtend.core.Xtend.Editor.opened">
-                        </reference>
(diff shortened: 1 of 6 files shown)


org.eclipse.xtend.ide/plugin.xml
```diff
@@ -1571,4 +1559,14 @@
      		 </adapt>
       </definition>
    </extension>
+   <extension
+         point="org.eclipse.core.runtime.adapters">
+      <factory
+            adaptableType="org.eclipse.xtext.xbase.ui.editor.XbaseEditor"
```
cdietrich: does this affect other editor like domain model?

kthoms: Good point. the action will also appear for dmodel, but with empty contents. On the other side there is a "Open Generated File" action which opens the generated Java code in an editor. I would expect that it should also use the DerivedSourceView. Maybe we should move it to Xbase?

cdietrich: If I remember correctly there is a bugzilla for that somewhere

cdietrich: https://bugs.eclipse.org/bugs/show_bug.cgi?id=381839

cdietrich: and drived resource view and ope generated file are not the same

kthoms: I guess generalizing this to Xbase is a bigger issue. As a workaround we could create an XtendEditor deriving from XbaseEditor to have a more concrete target for now. What do you think?

cdietrich: :+1:

kthoms:
Updated. The action is now only available for Xtend editors.

cdietrich:
Lgtm

Issue #134, opened 2017-01-19
Broken indent on create field in inner class


tivervac:
When inserting a field through the quickfix, in an inner class, the wrong indent is used and existing indent is broken.
```
class Foo {
	private static class Bar {

		new() {
			this.foobar = 0
		}
	}
}
```

Creating a field for foobar through the quickfix yields:
```
class Foo {
	private static class Bar {

	int foobar
	
	new() {
			this.foobar = 0
		}
	}
}
```

cdietrich:
all quickfixes are affected by this issues cause they use hard coded indentation levels

tivervac:
@cdietrich as your branch has been merged, can this be closed now?

cdietrich:
yes thx for the hint: fixed in Xtext/Xtend 2.13

PR #264, opened 2017-09-20
Broken indent on create field in inner class #134

cdietrich:
Broken indent on create field in inner class #134


cdietrich:
please review

PR #288, opened 2017-10-22
Change the visibility of isRelevantAnnotationTarget.

gallandarakhneorg:
In order to use the isRelevantAnnotationTarget() function within subtypes of XtendValidator, the visibility of this function should be changed from "private" to "protected".


cdietrich:
thx

Issue #290, opened 2017-10-24
Unnecessary proxy resolution in xtend compiler causes trouble with Scala classes


janhicken:
In 2nd transitive level, a class extending the scala class `akka.actor.UntypedActor` is used. When compiling the xtend file, the following error pops up:

```
[ERROR] resource is empty: java:/Objects/scala.collection.mutable.IndexedSeqView
java.lang.IllegalStateException
	at org.eclipse.xtext.common.types.access.impl.AbstractClassMirror.getEObject(AbstractClassMirror.java:94)
	at org.eclipse.xtext.common.types.access.TypeResource.getEObject(TypeResource.java:90)
	at org.eclipse.emf.ecore.resource.impl.ResourceSetImpl.getEObject(ResourceSetImpl.java:223)
	at org.eclipse.emf.ecore.util.EcoreUtil.resolve(EcoreUtil.java:203)
	at org.eclipse.emf.ecore.util.EcoreUtil.resolve(EcoreUtil.java:259)
	at org.eclipse.xtext.common.types.access.TypeResource.resolveJavaObjectURIProxy(TypeResource.java:172)
	at org.eclipse.xtext.common.types.impl.JvmTypeReferenceImplCustom.eResolveProxy(JvmTypeReferenceImplCustom.java:27)
	at org.eclipse.xtext.common.types.impl.JvmParameterizedTypeReferenceImpl.getType(JvmParameterizedTypeReferenceImpl.java:111)
	at org.eclipse.xtext.common.types.impl.JvmParameterizedTypeReferenceImpl.eGet(JvmParameterizedTypeReferenceImpl.java:173)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eGet(BasicEObjectImpl.java:1011)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eGet(BasicEObjectImpl.java:1003)
	at org.eclipse.emf.ecore.util.EContentsEList$FeatureIteratorImpl.hasNext(EContentsEList.java:439)
	at org.eclipse.emf.ecore.util.EcoreUtil.resolveCrossReferences(EcoreUtil.java:304)
	at org.eclipse.emf.ecore.util.EcoreUtil.resolveAll(EcoreUtil.java:298)
	at org.eclipse.emf.ecore.util.EcoreUtil.resolveAll(EcoreUtil.java:283)
[...]

PR #291, opened 2017-10-26
dont resolve and validate type resources #290

cdietrich:

### DIFF
--- a/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
+++ b/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
@@ -1,12 +1,20 @@
 package org.eclipse.xtend.core.compiler.batch;
 
 import static com.google.common.collect.Iterables.*;
+import static com.google.common.collect.Iterables.addAll;
+import static com.google.common.collect.Iterables.concat;
+import static com.google.common.collect.Iterables.filter;
+import static com.google.common.collect.Iterables.isEmpty;
+import static com.google.common.collect.Iterables.transform;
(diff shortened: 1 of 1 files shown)


org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
```diff
@@ -662,6 +672,9 @@ public void close() throws IOException {
 		List<Issue> issues = Lists.newArrayList();
 		List<Resource> resources = Lists.newArrayList(resourceSet.getResources());
 		for (Resource resource : resources) {
+			if (resource instanceof TypeResource) {
```
szarnekow: We could use isSourceFile(..) here, instead and remove the second half of the guard below.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
```diff
@@ -486,7 +487,16 @@ public boolean compile() {
 			}
 			// install a fresh type provider for the second phase, so we clear all previously cached classes and misses.
 			installJvmTypeProvider(resourceSet, classDirectory, false);
-			EcoreUtil.resolveAll(resourceSet);
+			List<Resource> toBeResolved = new ArrayList<>(resourceSet.getResources().size());
+			for (Resource resource : resourceSet.getResources()) {
+				if (resource instanceof TypeResource) {
```
szarnekow: Should we use isSourceFile here, too?

cdietrich:
good catches @szarnekow. updated the pr. validate has already the guard

PR #292, opened 2017-10-27
update to 2.14

cdietrich:

### DIFF
--- a/gradle/bootstrap-setup.gradle
+++ b/gradle/bootstrap-setup.gradle
@@ -12,7 +12,7 @@ repositories {
 }
 
 // The Xtend compiler version to use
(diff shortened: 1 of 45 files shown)


gradle/bootstrap-setup.gradle
```diff
@@ -12,7 +12,7 @@ repositories {
 }
 
 // The Xtend compiler version to use
-def bootstrapXtendVersion = '2.12.0.M1'
+def bootstrapXtendVersion = '2.13.0'
```
szarnekow: Is it possible to use smth like 2.14.0-snapshot here (as soon as there is a snapshot), so we'll get earlier feedback?

cdietrich: will have to have a look

PR #299, opened 2017-11-10
[455102] made getClassloader public as it is public in subclasses anyway

cdietrich:
[455102] made getClassloader public as it is public in subclasses anyway


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=455102

PR #312, opened 2017-11-28
moved idea to xtext-idea

cdietrich:


kthoms:
+1
Code should be maintained along with generic IDEA support. Let's keep the Xtend codebase clean from that. I think this should be decided on Gitter.

spoenemann:
See eclipse/xtext-idea#53.

Why should we move the code? If it's no longer supported, we can as well just delete it, and retrieve it from the history in the unlikely case that someone wants to restart working on it in the future.

kthoms:
Since there is no veto against this after addressing it on Gitter, this move can be performed.

PR #320, opened 2017-12-07
Fixed problem with Target Platforms not being properly installed

cdietrich:
Fixed problem with Target Platforms not being properly installed
https://github.com/eclipse/xtext-eclipse/issues/472

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,9 @@
 node {
 	properties([
-		[$class: 'BuildDiscarderProperty', strategy: [$class: 'LogRotator', numToKeepStr: '15']]
+		[$class: 'BuildDiscarderProperty', strategy: [$class: 'LogRotator', numToKeepStr: '15']],
+		parameters([
+			choice(choices: 'luna\noxygen\nphoton', description: 'Which Target Platform should be used?', name: 'target_platform')
+		])
 	])
 	
 	stage('Checkout') {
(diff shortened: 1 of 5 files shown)


releng/org.eclipse.xtend.target/pom.xml
```diff
@@ -28,6 +28,11 @@
 						<configuration>
 							<artifacts>
 								<artifact>
+                                    <file>${project.artifactId}.target</file>
```
szarnekow: Formatting is messed up?

cdietrich:
@kthoms ping

Jenkinsfile
```diff
@@ -23,6 +26,12 @@ node {
...
 	
 	stage('Maven Tycho Build') {
+		def targetProfile = "-Pluna"
+		if ("oxygen" == params.target_platform) {
+			targetProfile = "-Poxygen"
+		} else if ("photon" == params.target_platform) {
+			targetProfile = "-Pphoton"
+		}
 		wrap([$class:'Xvnc', useXauthority: true]) {
 			sh "${mvnHome}/bin/mvn -f tycho-pom.xml --batch-mode -fae -Dmaven.test.failure.ignore=true -Dmaven.repo.local=${workspace}/.m2/repository -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn clean install"
```
kthoms: `${targetProfile}` argument is missing

Jenkinsfile
```diff
@@ -23,6 +26,12 @@ node {
 	}
 	
 	stage('Maven Tycho Build') {
+		def targetProfile = "-Pluna"
+		if ("oxygen" == params.target_platform) {
+			targetProfile = "-Poxygen"
```
kthoms: I think `luna` has to be deactivated, since the profile is active by default. It would work also that way (only oxygen/photon is used as target), but more or less by accident.

```
-P-luna,+oxygen
```

Jenkinsfile
cdietrich: ?????

Jenkinsfile
cdietrich: nope pleae test

Jenkinsfile
cdietrich: got it

Issue #314, opened 2017-11-30
Don't create Deprecation issues on imports if the top-level type is deprecated.


szarnekow:
See behaviour of the JDT.

cdietrich:
@szarnekow which type do you look at if there are multiple

szarnekow:
To be discussed. We could either use the first type, consider all types (at least one is `@Deprecated` or check for the usages of the imports and look at the containing types.

cdietrich:
We Decided to use the "any top level type is deprecated" implementation for this issue

cdietrich:
Fixed in 2.14

PR #321, opened 2017-12-08
Don't create Deprecation issues on imports if the top-level type is deprecated. #314

cdietrich:
Don't create Deprecation issues on imports if the top-level type is deprecated. #314

### DIFF
--- a/org.eclipse.xtend.core.tests/testdata/testdata/DeprecatedClass.java
+++ b/org.eclipse.xtend.core.tests/testdata/testdata/DeprecatedClass.java
@@ -1,0 +1,16 @@
+/*******************************************************************************
+ * Copyright (c) 2017 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package testdata;
+
+/**
+ * @author Christian Dietrich - Initial contribution and API
+ */
+@Deprecated
+public class DeprecatedClass {
+
+}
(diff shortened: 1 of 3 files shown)


cdietrich:
please review

org.eclipse.xtend.core.tests/testdata/testdata/DeprecatedClass.java
```diff
@@ -0,0 +1,16 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package testdata;
+
+/**
+ * @author Christian Dietrich - Initial contribution and API
+ */
+@Deprecated
+public class DeprecatedClass {
```
szarnekow: Would it make sense to use org.eclipse.xtext.common.types.util.Deprecation2 instead?

cdietrich: is not visible

org.eclipse.xtend.core/src/org/eclipse/xtend/core/validation/XtendValidator.java
```diff
@@ -2124,4 +2125,18 @@ public void accept(XExpression implicitReturn) {
 	protected boolean isLocalClassSemantics(EObject object) {
 		return super.isLocalClassSemantics(object) || (object instanceof XtendMember && !(object instanceof AnonymousClass));
 	}
+	
+	@Override
+	public void checkDeprecated(XImportDeclaration decl) {
+		XtendFile file = EcoreUtil2.getContainerOfType(decl, XtendFile.class);
+		if (file != null) {
+			for (XtendTypeDeclaration t : file.getXtendTypes()) {
+				if (hasAnnotation(t, Deprecated.class)) {
```
szarnekow: Maybe check on the jvm level (`org.eclipse.xtext.common.types.util.DeprecationUtil.isDeprecated(JvmAnnotationTarget)`) so that an active annotation can also mark things as deprecated.

cdietrich:
updated. please have a look again

PR #325, opened 2017-12-13
increased maven/gradle asm version to 6.0

cdietrich:
increased maven/gradle asm version to 6.0
https://github.com/eclipse/xtext-extras/issues/208


kthoms:
First build was failing. I have retriggered it.

PR #326, opened 2017-12-13
[J9] consume jdt version in maven/gradle that does support Java 9

cdietrich:
[J9] consume jdt version in maven/gradle that does support Java 9
https://github.com/eclipse/xtext-extras/issues/211


org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
```diff
@@ -113,6 +113,17 @@ public boolean apply(Issue issue) {
 
 	private static Logger log = Logger.getLogger(XtendBatchCompiler.class.getName());
 
+	private static final boolean isJava9CompilerApiAvailable;
+	static {
+		boolean result = true;
+		try {
+			Class.forName("javax.lang.model.element.ModuleElement");
+		} catch (ClassNotFoundException e) {
```
kthoms: you could assign `isJava9CompilerApiAvailable=true` in the try clause and remove the result variable

cdietrich: code will be deleted as soon as oxygen2 is available

kthoms: so just wait until tomorrow?

cdietrich: or january, dont know when theyll publish to maven

kthoms: so temporary commit to make it work meanwhile? maybe if thursday JDT is not available on central?

cdietrich: no ill wait

gradle/versions.gradle
```diff
@@ -8,8 +8,9 @@ ext.versions = [
 	'xtext': version,
 	'xtext_gradle_plugin': '1.0.20',
 	'idea' : project.hasProperty('ideaEAP') ? 'LATEST-EAP-SNAPSHOT' : '145.972.3',
-	'jdtCore': '3.12.2',
-	'jdtCompilerApt': '1.2.100',
+	'jdtCore': '3.13.100',
+	'jdtCompilerApt': '1.3.60',
+	'jdtCompilerTool': '1.2.51',
 	'emfMwe2': '2.9.1.201705291010',
```
szarnekow: Should we use a more recent mwe2 version

cdietrich: there is no newer mwe yet

PR #338, opened 2018-01-09
[529291] Replace @build@ token in about.mappings

kthoms:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,6 +42,6 @@ subprojects {
 
 task clean(type: Delete) {
 	group 'Build'
-	description 'Deletes the local repositories'
+	description 'Deletes the local repositories.'
 	delete 'build'
 }
--- a/gradle/eclipse-project-layout.gradle
+++ b/gradle/eclipse-project-layout.gradle
@@ -1,3 +1,4 @@
+import org.apache.tools.ant.filters.*
 /*
  * Since we use the Eclipse Style layout where sources and resources live in the same
  * folders, we need to make some adjustments to Gradle's defaults.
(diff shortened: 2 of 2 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=529291

build.gradle
```diff
@@ -19,6 +19,7 @@ apply from: "${rootDir}/gradle/versions.gradle"
 apply from: "${rootDir}/gradle/bootstrap-setup.gradle"
 
 ext.buildTime = DateTimeFormatter.ofPattern('yyyyMMdd-HHmm').format(LocalDateTime.now())
+ext.buildType = "I" // I=Integration,N=Nightly,R=Release
```
cdietrich: how will this be passed in jenkins?

kthoms: `gradle -PbuildType=R ...`

cdietrich: maybe we should introduce params like we did for luna,oxygen,photon
@spoenemann any comments?

spoenemann:
> maybe we should introduce params like we did for luna,oxygen,photon

What do you mean?

In upstream-repositories.gradle we detect the type of build based on the version scheme. Couldn't we do the same here? It would be great to avoid another field that has to be set manually for every release.

cdietrich:
no i mean in the relase process one would have to adapt the jenkins file. is that ok for you? or should we introduce params like at http://services.typefox.io/open-source/jenkins/job/xtext-xtend/job/master/build?delay=0sec

kthoms:
I'll try to reuse the build type detection.

build.gradle
```diff
@@ -42,6 +42,34 @@ subprojects {
...
+		return 'R'
+}
+
+/**
+ * Computes a build identifier as a combination of the build type
+ * (Integration,Milestone/Release) and the build timestamp.
+ */
+def getBuildId () {
+	return getBuildType()+buildTime
+}
```
spoenemann: Since the build type and build id are only related to Eclipse and not to the Gradle/Maven artifacts, this code could be moved to eclipse-project-layout.gradle where it is used?

kthoms: Yes, this would be an option.

kthoms: updated.

Issue #340, opened 2018-01-11
DirtyStateEditorValidationTest.testBug410575 fails


ArneDeutsch:
The test case DirtyStateEditorValidationTest.testBug410575 fails when executed from maven but not when run inside eclipse. Investigate and fix.

cdietrich:
Fixed in 2.14 thanks @ArneDeutsch

PR #341, opened 2018-01-11
[#340] Fix test by ensure that preconditions are met.

ArneDeutsch:
The test validates that the editors synchronize between one another even
if the file is not saved. But it has to assume that the build is run
before the editors are opened. Seems on CLI the autobuild is not run
fast enough and in result the "waitForAutobuild" is passed before the
build has even started. But this autobuild is not the intention of the
test, it is part of the setup. Hence we just initiate a build manually
just as the other test cases do, too. And wait for this build to finish.
Afterwards the testss succeed on CLI as well as in eclipse.

Change-Id: I13b8362a7fca8d54d532a1b789f1a3456a8a042f


cdietrich:
@szarnekow i analyzed this together with @ArneDeutsch so i want a second opinion by you

cdietrich:
Discussed with @szarnekow in person. its ok to do the explicit build

PR #354, opened 2018-02-05
set lower guava bound to < 22

cdietrich:


szarnekow:
Should we set it to 24 or 25 instead? After all, Guava 24.0 was just released (https://github.com/google/guava/releases/tag/v24.0)

cdietrich:
this changes is only intended to be onpar with tycho metadata

Issue #358, opened 2018-02-09
Xtend Batch Compiler Main waits for ThreadExecutor


cdietrich:
Xtend Batch Compiler Main waits for the ThreadExecutor created by AnnotationProcessor
I propose to exit the main with {{System.exit(0)}} or to offer a second main that does so.
Or to explicitely shut the executor down (question would be how to access it)

cdietrich:
[Main-2018-02-09-shutdown.snapshot.zip](https://github.com/eclipse/xtext-xtend/files/1711020/Main-2018-02-09-shutdown.snapshot.zip)

cdietrich:
can be reproduced by calling `org.eclipse.xtend.core.compiler.batch.Main` with
```
-d /data/eclipse-workspaces/ws-dev-cdietric-largedevices/demo/src-gen2 -cp /home/cdietric/.p2/pool/plugins/com.google.guava_15.0.0.v201403281430.jar:/home/cdietric/.p2/pool/plugins/org.eclipse.xtext.xbase.lib_2.13.0.v20171012-0742.jar:/home/cdietric/.p2/pool/plugins/org.eclipse.xtend.lib_2.13.0.v20171012-0742.jar:/home/cdietric/.p2/pool/plugins/org.eclipse.xtend.lib.macro_2.13.0.v20171012-0742.jar /data/eclipse-workspaces/ws-dev-cdietric-largedevices/demo/src
``` 

on
```
package demo

import org.eclipse.xtend.lib.annotations.Accessors

class Demo {
    
    @Accessors
    var String x
    
}
```

cdietrich:
possible solutions:

(1) shut down executor service
(2) system.exit(0)

cdietrich:
Fixed in 2.14

PR #359, opened 2018-02-09
[#358] shutdown executor service after compile

cdietrich:
[#358] shutdown executor service after compile

### DIFF
--- a/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/Main.java
+++ b/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/Main.java
@@ -16,6 +16,7 @@ import java.util.List;
 
 import org.apache.log4j.BasicConfigurator;
 import org.eclipse.xtend.core.XtendInjectorSingleton;
+import org.eclipse.xtend.core.macro.AnnotationProcessor.CancellationObserver;
 
 import com.google.common.base.Joiner;
 import com.google.inject.Injector;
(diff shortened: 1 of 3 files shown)


org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/Main.java
```diff
@@ -68,6 +69,7 @@ public static void main(String[] args) {
 		if (!compiler.compile()) {
 			System.exit(1);
 		}
+		injector.getInstance(CancellationObserver.class).getPool().shutdownNow();
```
szarnekow: I'd prefer `CancellationObserver.stop()`

szarnekow: Also I'd prefer this to be internal to the XtendBatchCompiler similarly to `destroyClassLoader` et al

org.eclipse.xtend.core/xtend-gen/org/eclipse/xtend/core/macro/AnnotationProcessor.java
```diff
@@ -62,6 +62,10 @@ protected ExecutorService initPool() {
       return Executors.newCachedThreadPool();
     }
     
+    public List<Runnable> stop() {
+      return this.pool.shutdownNow();
```
szarnekow: please use `shutdown()` instead of `shutdownNow()` for orderly shutdown.

Issue #361, opened 2018-02-13
Implement Xtend Validation rule "Redundant modifiers"


miklossy:
In Xtend, the usage of the 'final' keyword in combination with the 'val' keyword is unnecessary. 

```Xtend
class A {
	final val a = 1
}
```
A Warning could be raised on the 'final' keyword in such cases.

cdietrich:
could be added to `org.eclipse.xtend.core.validation.ModifierValidator.checkModifiers(XtendMember, String)`

kthoms:
@miklossy Could you also provide a PR for a quickfix? I have opened #370 for that.

kthoms:
How about making it more generic? The general issue is an unnecessary modifier. This is also true for public on classes or private on fields.

miklossy:
@cdietrich I think this issue could be closed, since the PR has already been merged.

miklossy:
The following cases have not been considered yet and should also be marked as 'Redundant modifier':

- public modifier on interfaces
- public modifier on constructors

PR #362, opened 2018-02-13
[#361] Implement "Redundant modifiers" Xtend validation rule.

miklossy:
Extend the Xtend modifier validator with "Redundant modifiers" to raise warnings in case of:
- 'final' in combination with 'val'
- 'private' on fields
- 'public' on methods
- 'public' on classes
- 'def' in combination with 'override'
- Implement corresponding ModifierValidationTest test cases.
- Eliminate redundant modifiers from the Xtend test data in order not to break existing test cases.

### DIFF
--- a/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/ModifierValidationTest.xtend
+++ b/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/ModifierValidationTest.xtend
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2013, 2017 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2013, 2018 itemis AG (http://www.itemis.eu) and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -31,6 +31,7 @@ class ModifierValidationTest extends AbstractXtendTestCase {
 		clazz('''abstract class Foo{}''').assertNoErrors
 		clazz('''dispatch class Foo{}''').assertError(XTEND_CLASS, INVALID_MODIFIER)
 		clazz('''final class Foo{}''').assertNoErrors
+		clazz('''public class Foo{}''').assertWarning(XTEND_CLASS, REDUNDANT_MODIFIER)
 	}
 	
 	@Test def void testNestedClassAllowedModifiers() {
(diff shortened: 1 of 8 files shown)


kthoms:
How about making it more generic? The general issue is an unnecessary modifier. This is also true for `public` on classes or `private`on fields.

The issue would be better named `REDUNDANT_MODIFIER` then.

miklossy:
I have just updated the commit: Implement Xtend Validation rule "Redundant modifiers". Warning will be raised in case of:
- 'final' in combination with 'val'
- 'private' on fields
- 'public' on methods
- 'public' on classes
- 'def' in combination with 'override'

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/ModifierValidationTest.xtend
```diff
@@ -129,6 +130,7 @@ class ModifierValidationTest extends AbstractXtendTestCase {
 		abstractFunction('''static def int foo();''').assertError(XTEND_FUNCTION, INVALID_MODIFIER)
 		function('''dispatch def foo (int i){}''').assertNoErrors
 		function('''final def foo() {}''').assertNoErrors
+		function('''public def foo() {}''').assertWarning(XTEND_FUNCTION, REDUNDANT_MODIFIER)
```
cdietrich: did you see the issue for a quickfix. maybe its better to add the redudant modifier as issue data so that it can be picked up by the quickfix. @szarnekow what do you think?

szarnekow: Yes, we should store all the redundant modifier indexes as user data on the issue.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/validation/ModifierValidator.java
```diff
@@ -96,16 +108,52 @@ protected void checkModifiers(XtendMember member, String memberName) {
...
+				defSeen = true;
 			}
 		}
 	}
+
+	/**
+	 * @since 2.14
+	 */
+	protected void redundantModifierWarning(String modifier, String memberName, EObject source, int index) {
+		warning("The "+ modifier + " modifier is redundant on " + memberName, source, index, IssueCodes.REDUNDANT_MODIFIER);
```
cdietrich: should we have a warniung for each and every separate redundant modifier or one for all

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/ModifierValidationTest.xtend
miklossy: Yes, I saw that ticket also. I think it would make sense to implement the quick fixes (and everything that servers that purpose) in a separate Pull Request.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/validation/ModifierValidator.java
miklossy: I think one warning for each separate redundant modifier is ok, the redundant modifier check will produce at most two warnings per Xtend Member.

cdietrich:
@kthoms @szarnekow anything left?

PR #368, opened 2018-02-18
Unit tests for unused field and local variables.

gallandarakhneorg:
The included unit tests are related to the change of the
"isLocallyUsed()" function within xbase-extras (PR 232).
Any field or local variable that is used as a left operand
of the assignment operator is not considered as being used.

see https://github.com/eclipse/xtext-extras/pull/232

### DIFF
--- a/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/ReadWriteAccessTest.java
+++ b/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/ReadWriteAccessTest.java
@@ -1,0 +1,639 @@
+/*******************************************************************************
+ * Copyright (c) 2018 Belfort-Montbeliard University of Technology (http://www.utbm.fr) and others.
+ * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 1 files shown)


org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/ReadWriteAccessTest.java
```diff
@@ -0,0 +1,639 @@
...
+ *******************************************************************************/
+package org.eclipse.xtend.core.tests.validation;
+
+import static org.eclipse.xtend.core.validation.IssueCodes.*;
+import static org.eclipse.xtend.core.xtend.XtendPackage.Literals.*;
+import static org.eclipse.xtext.xbase.XbasePackage.Literals.*;
+import static org.eclipse.xtext.xbase.validation.IssueCodes.*;
+import static org.eclipse.xtext.xtype.XtypePackage.Literals.*;
+
+import javax.inject.Inject;
```
szarnekow: Should be com.google.inject

PR #369, opened 2018-02-20
use junit 4.12

cdietrich:


cdietrich:
got somehow lost at @ArneDeutsch an my communication

Issue #370, opened 2018-02-20
Provide quickfix to remove redundant modifiers


kthoms:
#361 introduces a validation rule for unnecessary use of `final` together with `val`. A quickfix should be offered to remove the modifier.

cdietrich:
cc @miklossy

cdietrich:
fixed in 2.14

cdietrich:
thx @miklossy

PR #387, opened 2018-03-19
[#370] Provider quickfixes to remove redundant Xtend member modifiers.

miklossy:
- Extend the ModifierValidator to pass the redundant modifier as issue
data from the validator to the quickfix provider.
- Extend the XtendQuickfixProvider to provide quickfixes to remove the
redundant Xtend member modifier.
- Implement corresponding Xtend QuickfixTest test cases.

### DIFF
--- a/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/quickfix/XtendQuickfixProvider.java
+++ b/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/quickfix/XtendQuickfixProvider.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2017 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2011, 2018 itemis AG (http://www.itemis.eu) and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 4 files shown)


org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/quickfix/XtendQuickfixProvider.java
```diff
@@ -649,6 +651,25 @@ public void apply(EObject element, IModificationContext context) throws Exceptio
...
+		String[] issueData = issue.getData();
+		if(issueData==null || issueData.length==0) {
+			return;
+		}
+		
+		String modifier = issueData[0];
+		String label = "Remove the '" + modifier + "' modifier.";
+		String description = "The modifier '" + modifier + "' is unnecessary and should be removed.";
+		
+		acceptor.acceptMulti(issue, label, description, "fix_indent.gif", new IMultiModification<XtendMember>() {
```
cdietrich: how does this behave if i quickfix multiple issues with different modifiers?
i believe the resultion should read the data out of the actual issue.
or it should filter issues with the same data

cdietrich: or should do the test for redundant inside the fix and remove all redundant ones

miklossy:
Thanks for the hint, @cdietrich ! I updated the commit in the PR so that the created issue resolutions use the same label, description and image. Now it is possible to execute the quickfixes in batch mode even if they will remove different modifiers.

PR #402, opened 2018-03-24
[460455] Add OrganizeImportsTest for avoiding fully qualifying names.

miklossy:
- Test case for https://github.com/eclipse/xtext-extras/pull/243
- Activate/Adapt OrganizeImportsTest test case to verify that the import
organizer does not fully qualify type names if two types (one as inner
type, one as top-level type) with the same name exists.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=460455

PR #411, opened 2018-04-02
[495314] Fixed stackoverflow in type computation

szarnekow:
Tests are here. Impl is in extras.


cdietrich:
see CommonSuperTypeTest

cdietrich:
restarting the build helped. repo split ftw

PR #422, opened 2018-04-16
[#361] 'Redundant modifier' validation rule on interfaces/constructors.

miklossy:
- Extend the Xtend ModifierValidator to recognize the 'public' modifier
on interfaces and constructors as 'Redundant modifier'.
- Implement corresponding QuickfixTest test cases.
- Adapt the input data of the already existing test cases in order not
to break them.

### DIFF
--- a/org.eclipse.xtend.core/src/org/eclipse/xtend/core/validation/ModifierValidator.java
+++ b/org.eclipse.xtend.core/src/org/eclipse/xtend/core/validation/ModifierValidator.java
@@ -15,9 +15,13 @@ import java.util.List;
 import java.util.Set;
 
 import org.eclipse.emf.ecore.EObject;
+import org.eclipse.xtend.core.xtend.XtendAnnotationType;
 import org.eclipse.xtend.core.xtend.XtendClass;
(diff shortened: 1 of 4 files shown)


org.eclipse.xtend.core/src/org/eclipse/xtend/core/validation/ModifierValidator.java
```diff
@@ -77,7 +79,12 @@ protected void checkModifiers(XtendMember member, String memberName) {
 					if("private".equals(modifier) && member instanceof XtendField) {
 						redundantModifierWarning("private", memberName, member, i);
 					}
-					if("public".equals(modifier) && (member instanceof XtendClass || member instanceof XtendFunction)) {
+					if("public".equals(modifier) && (
+							member instanceof XtendClass ||
+							member instanceof XtendInterface ||
```
cdietrich: what about enums and annotations?

miklossy:
Thanks for the hint @cdietrich!

I updated the commit in the PR so that annotations and enums with 'public' modifiers are also recognized.

cdietrich:
Thx @miklossy

PR #423, opened 2018-04-18
fixed broken launch config

cdietrich:


ArneDeutsch:
Code change is ok.
Failing build is not related to the change.

PR #424, opened 2018-04-18
[#361] Rename the 'Redundant modifier' to 'Unnecessary modifier'.

miklossy:
- Since the validation rule mostly targets visibility modifiers matching
the defaults, using the 'Unnecessary modifier' notation is more suitable
than the 'Redundant modifier' notation.


cdietrich:
Thx @miklossy

PR #425, opened 2018-04-20
Java 10 Integration

cdietrich:
use asm 6.1.1
use jdt oxygen
adjusted pipelines


org.eclipse.xtend.caliper.tests/META-INF/MANIFEST.MF
```diff
@@ -16,7 +16,7 @@ Require-Bundle: com.google.inject;bundle-version="3.0.0",
  org.eclipse.xtext.xbase.testing,
  org.eclipse.jdt.core;bundle-version="3.6.0",
  org.eclipse.xtend.ide,
- org.objectweb.asm;bundle-version="5.0.1"
+ org.objectweb.asm;bundle-version="[6.1.1,7.0.0)"
```
holgerschill: Why upper bound to 7.0.0? Why not 6.1.1 as fixed version?

cdietrich: to be more flexible. at other places we have and had already a range [5.0.1,7.0.0)

PR #433, opened 2018-04-24
[eclipse/xtext-core#710] Workaround for exception.

ArneDeutsch:
Implement a workaround to avoid exception. Formatting is not 100%
correct but at least the formatting is done at all (one missing tab).

Added test case to verify that no exception is raised.

Also added TODO comments explaining what and why it was done.


ArneDeutsch:
Right now the test fails. Seems it is because the code is formatted and the formatter now indents the empty line inside the `<<IF>>` with three tabs. @miklossy I have added the three tabs to make the test succeed. Would you please review if these modification still is valid from your perspective?

miklossy:
I agree with your modification. Thx @ArneDeutsch!

cdietrich:
http://services.typefox.io/open-source/jenkins/job/xtext-xtend/job/arne-xtext-core-710/

PR #467, opened 2018-06-11
[eclipse/xtext-core#710] Workaround for exception.

ArneDeutsch:
Implement a workaround to avoid exception. Formatting is not 100%
correct but at least the formatting is done at all (one missing tab).

Added test case to verify that no exception is raised.

Also added TODO comments explaining what and why it was done.

Enabled and fixed added QuickfixTest.


cdietrich:
@ArneDeutsch does this replace https://github.com/eclipse/xtext-xtend/pull/433

ArneDeutsch:
Yes ... not sure why I have created two merge requests. They are identical.

Issue #468, opened 2018-06-13
Add Thread Looks to maven Plugins


cdietrich:
we should add the maven threading lock
https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3
to our xtend maven plugins too
https://github.com/eclipse/xtext-maven/issues/36

cdietrich:
Fixed in 2.15

PR #470, opened 2018-06-15
[#468] Marked Mojos as Threadsafe and introduced a lock to ensure singlethreaded execution

cdietrich:
[#468] Marked Mojos as Threadsafe and introduced a lock to ensure singlethreaded execution

### DIFF
--- a/org.eclipse.xtend.maven.plugin/src/main/java/org/eclipse/xtend/maven/AbstractXtendMojo.java
+++ b/org.eclipse.xtend.maven.plugin/src/main/java/org/eclipse/xtend/maven/AbstractXtendMojo.java
@@ -9,6 +9,8 @@ import com.google.inject.Inject;
 
 public abstract class AbstractXtendMojo extends AbstractMojo {
 
+	private static final Object lock = new Object();
+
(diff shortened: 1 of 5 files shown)


cdietrich:
please review

org.eclipse.xtend.maven.plugin/src/main/java/org/eclipse/xtend/maven/AbstractXtendMojo.java
```diff
@@ -9,6 +9,8 @@
 
 public abstract class AbstractXtendMojo extends AbstractMojo {
 
+	private static final Object lock = new Object();
```
szarnekow: Don't they use separate class loaders?

cdietrich: does not look like. here is the pattern to avoid multitreaded execution: https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3 i basically did what they proposed

PR #482, opened 2018-06-26
[402920] Add QuickfixTest to ensure the creation of Xtend notation.

miklossy:
- Implement useXtendTypeSignature Quickfix test case to ensure that the
Quickfix creates the missing method signature in the Xtend notation
instead of using the Java notation.
- Adjust the QuickfixTestBuilder test helper class to consider only the
first '|' symbol as a position marker to be able to use further '|'
symbols in Xtend lambda expressions.

### DIFF
--- a/org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/QuickfixTestBuilder.xtend
+++ b/org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/QuickfixTestBuilder.xtend
@@ -3,24 +3,25 @@ package org.eclipse.xtend.ide.tests.quickfix
 import com.google.inject.Inject
 import java.util.List
 import java.util.Set
+import java.util.regex.Pattern
 import org.eclipse.core.resources.IFile
 import org.eclipse.core.runtime.NullProgressMonitor
 import org.eclipse.jface.preference.IPersistentPreferenceStore
 import org.eclipse.xtend.ide.tests.WorkbenchTestHelper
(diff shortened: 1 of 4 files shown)


szarnekow:
It's just test infra, but what do you think about this idea:
- Attempt 1: try to replace the position marker <|> - if there is not exactly one, continue with attempt (2)
- Attempt 2: try to replace the position marker | - if there is not exactly one, fail the test with a hint that <|> may be used to disambiguate ordinary | from a position marker
- Proceed with the test execution

miklossy:
I like your idea @szarnekow, I have updated the PR accordingly.

org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/QuickfixTestBuilder.xtend
```diff
@@ -59,16 +63,16 @@ class QuickfixTestBuilder {
 		xbaseBuilderPreferenceAccess.setJavaVersion(project, javaVersion);
 	}
 	
-	def create(String fileName, CharSequence model) {
-		assertNotSame('No position marker | found in model', -1, model.toString.indexOf("|"))
-		val file = createFile(fileName, model.toString.replace("|", ""))
+	def create(String fileName, String model) {
+		val positionMarker = model.positionMarker
```
szarnekow: If (and only if) this would be a critical path, I'd rather go for a model.replaceAll() and compare the length of the result with the original length. That way you would immediately figure out, how many markers there are.
Nevertheless, the PR looks good to me :)

cdietrich:
Thx @miklossy

PR #491, opened 2018-06-30
fix #484: @Override for overridden dispatch methods

### LINKED ISSUE #484: overridden dispatch methods don't have @Override annotation
1. Have a method "protected def dispatch void m()" in AbstractSomething.xtend
2. Have a method "override dispatch void m()" in Something.xtend (extending AbstractSomething)

The generated method in Something.java has no @Override annotation (while other overridden methods generally have it).

Bananeweizen:
Signed-off-by: Michael Keppler <Michael.Keppler@gmx.de>


org.eclipse.xtend.core.tests/xtend-gen/org/eclipse/xtend/core/tests/compiler/CompilerBugDispatchWithOverrideTest.java
```diff
@@ -0,0 +1,191 @@
+/**
+ * Copyright (c) 2017 itemis AG (http://www.itemis.eu) and others.
```
cdietrich: should be 2018

org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/XtendJvmModelInferrer.java
```diff
@@ -620,18 +625,21 @@ protected void transform(XtendFunction source, JvmGenericType container, boolean
...
+			setOverride(operation);
 		if (createExtensionInfo != null) {
 			transformCreateExtension(source, createExtensionInfo, container, operation, returnType);
 		} else {
 			setBody(operation, expression);
 		}
 		jvmTypesBuilder.copyDocumentationTo(source, operation);
 	}
+
+	private void setOverride(JvmOperation operation) {
```
cdietrich: should be protected

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/CompilerBugDispatchWithOverrideTest.xtend
```diff
@@ -0,0 +1,104 @@
+/*******************************************************************************
+ * Copyright (c) 2017 itemis AG (http://www.itemis.eu) and others.
```
cdietrich: should be 2018

Bananeweizen: all fixed. Do you prefer new commits for such fixes or overwriting existing commits (as I just did)?

cdietrich: in this case i prefer a amend

cdietrich:
@szarnekow i would appreciate your review

org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/XtendJvmModelInferrer.java
```diff
@@ -536,6 +537,8 @@ protected JvmOperation deriveGenericDispatchOperationSignature(Iterable<JvmOpera
 				associator.associate(func, result);
 				if (!func.isStatic())
 					allStatic = false;
+				if (func.isOverride())
```
szarnekow: Unfortunately this won't be sufficient. I suspect that a case like
```
class Super {
  def dispatch void m(String s) {}
  def dispatch void m(Object s) {}
}
class Sub {
  def dispatch void m(Number n) {}
}
```
will miss the override annotation on the dispatcher.

szarnekow:
> i would appreciate your review
Sorry, that slipped through.

Issue #504, opened 2018-07-11
Provide Quickfixes to change the return type to void in JUnit methods.


miklossy:
[439703] introduces a validation rule to ensure that an Xtend method has a void return type if a JUnit annotation is active on it. Quickfixes should be offered in such cases as well.

cdietrich:
Fixed in 2.15

PR #505, opened 2018-07-11
[#504] Provide Quickfixes to change JUnit methods' return type to void.

miklossy:
- Extend the XtendQuickfixProvider to provide Quickfixes to change the
return type of an Xtend method to void (also executable in batch mode).
- Implement corresponding Xtend QuickfixTest test cases.


cdietrich:
i wonder what we should do if the method body actually contains return statements.
@szarnekow @kthoms what do you think?

cdietrich:
thx @miklossy

Issue #516, opened 2018-07-26
Improvements to 'Validation rule to ensure void return type in JUnit method'.


miklossy:
The [Pull Request](https://github.com/eclipse/xtext-xtend/pull/488) introduced a new Xtend validation rule to ensure the void return type on methods having JUnit annotations. 

I observed that such validation issues are also reported as follow up issues of other issues:
![screenshot](https://user-images.githubusercontent.com/20393472/43255437-1bedc3c0-90ca-11e8-84da-b85c561c861e.png)

I think it would be useful not to report such follow up validation issues, since the actual cause is somewhere else. What do you think? Would it be possible? / Does it makes sense at all?

szarnekow:
The validation should not issue errors if the method body does already contain errors I suppose.

szarnekow:
There is no infrastructure yet available to make `@Check` based validation dependent on each other or on dependent on the presence or absence of linking or syntax problems. I'd rather postpone this one.

cdietrich:
TODO:

- debug if the method is typed as "unknown"
- if yes change the check to not flag unknow types as non void

miklossy:
Thx for the nice suggession @cdietrich! I created the corresponding PR.

cdietrich:
Fixed in 2.15 thx for the patch @miklossy

dhuebner:
@miklossy @cdietrich 
Is the validation severity configurable? If so, in which errors/warnings section. 
Also, the validation should check the Test class `RunWith` annotation. A custom test runner can work with test method return types, in fact the validation prevents using such kinds of custom test runners in Xtend.

cdietrich:
Please open a new issue and fix it

dhuebner:
Well, it doesn't seem to be an issue for you.

cdietrich:
well you did no care the last 12 months so what

dhuebner:
Nice support, thanks!

cdietrich:
@dhuebner you provide more committers to Xtext than itemis does. so if you find something improvable so report it and provide a patch instead of reopening random old tickets and just complaining

miklossy:
I am absolutely agree with @cdietrich

PR #518, opened 2018-07-30
[#516] Improve the checkJUnitMethodReturnType Xtend validation rule.

miklossy:
- Modify the checkJUnitMethodReturnType Xtend validation rule to ensure
that the 'JUnit Method Return Type Validation Check' doesn't report any
follow up issues in case of an unknown return type.
- Implement corresponding JUnitMethodReturnTypeValidationTest test case.


org.eclipse.xtend.core/src/org/eclipse/xtend/core/validation/XtendValidator.java
```diff
@@ -341,7 +342,7 @@ public void checkJUnitMethodReturnType(XtendFunction function) {
 		 */
 		if(hasJUnitAnnotation(operation)) {
 			LightweightTypeReference actualType = determineReturnType(operation);
-			if(actualType !=null && !actualType.isPrimitiveVoid()) {
+			if(actualType !=null && !(actualType instanceof UnknownTypeReference) && !actualType.isPrimitiveVoid()) {
```
cdietrich: maybe simply calling isUnknown() on the actual type is better readable

cdietrich: @szarnekow what do you think?

miklossy:
Calling isUnknown() is indeed more readable. I updated the PR.

PR #526, opened 2018-08-03
allow interface to extend an annotation

cdietrich:


cdietrich:
updated test and impl

Issue #536, opened 2018-08-22
Organize import adds useless import


tivervac:
When you have code that has a static extension import, and you use that imported function in a static AND an extended way, an extra useless import is added.

```
package pkg;

import org.eclipse.emf.ecore.EObject

import static extension org.eclipse.emf.ecore.util.EcoreUtil.getURI

class Test {
	
	def void doSomething() {
		val EObject e = null
		val a = e.URI // Extended call
		val b = getURI(e) // Static call
	}
}
```
Organize import on the above code returns:
```
package pkg;

import org.eclipse.emf.ecore.EObject

import static org.eclipse.emf.ecore.util.EcoreUtil.* // This is useless

import static extension org.eclipse.emf.ecore.util.EcoreUtil.getURI

class Test {
	
	def void doSomething() {
		val EObject e = null
		val a = e.URI // Extended call
		val b = getURI(e) // Static call
	}
}
```

ArneDeutsch:
Added a check if the static import is already covered by a static extension import. If so, do not add the static import for that type.

cdietrich:
Fixed in 2.15. thx @ArneDeutsch

PR #542, opened 2018-08-26
[eclipse/xtext-xtend#536] Add test cases for new cases.

ArneDeutsch:


cdietrich:
@ArneDeutsch did you add upstream to gradle and target to see this does not break any other tests?

PR #551, opened 2018-09-01
Fixed two issues with Xtend File Hyperlinks in Consoles

cdietrich:
- [538503] catch NumberFormatExceptions that can be caused by changed console content
- Fixed problem with non launch based stacktraces


tivervac:
Pretty small commit LGTM, but are we curing the symptoms or the root cause?  @cdietrich

cdietrich:
We catch already the badlocation exception thus is just a generalization of the current thing. I don’t see a easy way how this could be fixed in eclipse ui

tivervac:
Yeah at that point we probably can't fix it anymore but it might be useful to check why there are numberformatexceptions or badlocationexceptions

cdietrich:
See the bugzilla. It’s caused by the document in the console is not read write threadsafe

tivervac:
I'll see if I can find the time

PR #553, opened 2018-09-02
Activate ignored Xtend Rename Refactoring test case.

miklossy:
- The following test case have been ignored, but now it is running
successfully (tested it also manually).


cdietrich:
thx @miklossy

PR #565, opened 2018-09-07
[#164] Add further test case reproducing the issue (currently ignored).

miklossy:
- Add testRenameXtendMember001 test case that is smaller and easier to
understand/debug than the testRenameXtendMember002 test case.


cdietrich:
thx @miklossy

PR #566, opened 2018-09-10
[eclipse/xtext#1224] Detect local Jenkins environment

kthoms:
Build steps defined in Jenkinsfile pass the built-in environment
variable 'JENKINS_URL' to the Gradle/Maven executions. This is evaluated
in the build scripts for upstream repository URLs. On Xtext JIPP this
will use upstream repos from JIPP. In local builds outside of Jenkins
the property defaults to Typefox CI like before.

Extend function jenkinsPipelineRepo() by upstreamBranch parameter and use this in the call.
Use jenkinsPipelineRepo() also for lsp4j.

### DIFF
--- a/gradle/bootstrap-setup.gradle
+++ b/gradle/bootstrap-setup.gradle
@@ -1,13 +1,16 @@
 /*
  * Root project configuration that is reused by subprojects to apply the Xtend compiler.
  */
+if (!hasProperty('JENKINS_URL')) {
+	ext.JENKINS_URL = 'http://services.typefox.io/open-source/jenkins'
+}
 
 // The repositories to query when constructing the Xtend compiler classpath
 repositories {
 	jcenter()
 	maven {
 		name 'xtend-bootstrap'
-		url 'http://services.typefox.io/open-source/jenkins/job/xtend-bootstrap/lastStableBuild/artifact/build-result/maven-repository/'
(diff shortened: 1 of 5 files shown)


gradle/bootstrap-setup.gradle
```diff
@@ -1,13 +1,16 @@
 /*
  * Root project configuration that is reused by subprojects to apply the Xtend compiler.
  */
+if (!hasProperty('JENKINS_URL')) {
```
cdietrich: is hasProperty = -P or -D i am confused now

kthoms: In Gradle it is "-P". In Maven "-D".

cdietrich: :+1:

kthoms:
The JENKINS_URL all-caps is a match to the build-in environment variable in Jenkins, see https://wiki.jenkins.io/display/JENKINS/Building+a+software+project#Buildingasoftwareproject-belowJenkinsSetEnvironmentVariables

> So gradle uses -P to pass properties whereas we have to use -D in mvn land?

Right. Gradle uses project properties with -P (https://docs.gradle.org/current/userguide/command_line_interface.html#sec:command_line_completion). Maven uses system properties with -D. Gradle also knows -D system properties, could be that this also works. But I think the right way is -P here.

kthoms:
Updated. URL for lsp4j stays hard-coded.

Issue #554, opened 2018-09-03
Generated Code has misordered modifiers 


hollesse:
If I use the method `addField()` from MutableClassDeclaration and want to make the field static and final these modifiers are misordered. 

Example code:
```
private def addTestField(MutableClassDeclaration class, extension TransformationContext context){
	class.addField("TEST", [
		type = boolean.newTypeReference
		final = true
		static = true
		visibility = Visibility::PRIVATE
		initializer = [ '''«false»''']
	])
}
```

Then the result is `private final static boolean TEST = false;` 
If you take a look at the java language spec it should be `private static final boolean TEST = false;`

cdietrich:
this is a general issue that is not tight to active annotation (see XtendGenerator.xtend)

miklossy:
We implemented the requested feature, the modifiers should be generated in the right order:  `static final` instead of `final static`.

@hollesse Would you please give it a try with the latest Xtend Version (nightly build) and give feedback, if the current behaviour corresponds to your expectations?

kthoms:
It is not necessary to use the Nightly build. It is contained in the 2.15.0 release version.

cdietrich:
no its not

cdietrich:
Fixed in 2.16

PR #577, opened 2018-09-17
[#554] Ensure static and final modifiers generated in the right order.

miklossy:
- Modify the XtendGenerator to generate the static and the final
modifiers in the right order (corresponding to the java language
specification) 'static final' instead of 'final static'.
- Adapt the corresponding test cases.


kthoms:
it is a recommendation only, but of course ok to follow it
spec: https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.1.1

PR #580, opened 2018-09-20
Activate useJenkinsSnapshots profile by default

kthoms:
Improves the experience with fresh checked out projects when dependent
snapshots are not available locally and must be downloaded.


cdietrich:
@kthoms no change for android archetype?

kthoms:
have to check, thanks for the hint

kthoms:
updated

kthoms:
yes we should do this.

kthoms:
the `org.eclipse.xtext.maven.parent` and `org.eclipse.xtext.maven.plugin` have this already set

PR #581, opened 2018-09-20
Replaced call to getComplianceLevel by getJavaSourceVersion

kthoms:

### DIFF
--- a/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
+++ b/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
@@ -636,7 +636,7 @@ public class XtendBatchCompiler {
 			commandLine.add("-cp \"" + Joiner.on(File.pathSeparator).join(classPathEntries) + "\"");
 		}
 		commandLine.add("-d \"" + classDirectory.toString() + "\"");
-		commandLine.add("-" + getComplianceLevel());
+		commandLine.add("-" + getJavaSourceVersion());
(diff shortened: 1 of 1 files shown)


org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
```diff
@@ -636,7 +636,7 @@ protected boolean preCompile(File classDirectory, Iterable<String> sourcePathDir
 			commandLine.add("-cp \"" + Joiner.on(File.pathSeparator).join(classPathEntries) + "\"");
 		}
 		commandLine.add("-d \"" + classDirectory.toString() + "\"");
-		commandLine.add("-" + getComplianceLevel());
+		commandLine.add("-" + getJavaSourceVersion());
```
cdietrich: we need to discuss this cause it breaks api

kthoms: In which regard? getComplianceLevel just call getJavaSourceVersion.
Or do you mean because they are semantically different?

cdietrich: it breaks subclasses overriding the old method. otherwise there was no reason to deprecate

kthoms: Ah, right. Who would do this? You'll never know...

kthoms:
OK, will abandon this one.

Issue #582, opened 2018-09-20
XtendBatchCompiler: -bootclasspath arg invalid on Java 9+


kthoms:
The XtendBatchCompiler does not consider the compliance level when adding `bootclasspath` to command line args. The argument is only allowed for Java 8 and older.

https://www.oracle.com/technetwork/java/javase/9-relnote-issues-3704069.html

cdietrich:
@kthoms is this fixed ?

kthoms:
Yes.

PR #585, opened 2018-09-21
[#582] Evaluate bootclasspath only for Java <= 1.8

kthoms:

### DIFF
--- a/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
+++ b/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
@@ -234,10 +234,17 @@ public class XtendBatchCompiler {
 	}
 
 	/**
+	 * This option is only supported on JDK 8 and older and will be ignored when source level is 9 or newer.
 	 * @since 2.7
+	 * @see https://www.oracle.com/technetwork/java/javase/9-relnote-issues-3704069.html
 	 */
 	public void setBootClassPath(String bootClassPath) {
-		this.bootClassPath = bootClassPath;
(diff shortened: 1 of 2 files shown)


org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
```diff
@@ -628,9 +628,14 @@ protected boolean preCompile(File classDirectory, Iterable<String> sourcePathDir
 		if (isVerbose()) {
 			commandLine.add("-verbose");
 		}
-		List<String> bootClassPathEntries = getBootClassPathEntries();
```
cdietrich: please add a javadoc hint to setbootclasspath
and we should discuss with what this needs to be replaced for java 9 and later

kthoms:
From org.eclipse.jdt.internal.compiler.batch.Main l.5250
```
protected long validateClasspathOptions(ArrayList<String> bootclasspaths, ArrayList<String> endorsedDirClasspaths, ArrayList<String> extdirsClasspaths) {
	if (this.complianceLevel > ClassFileConstants.JDK1_8) {
		if (bootclasspaths != null && bootclasspaths.size() > 0)
			throw new IllegalArgumentException(
				this.bind("configure.unsupportedOption", "-bootclasspath")); //$NON-NLS-1$ //$NON-NLS-2$
```

The structure from JDK 9+ differs from JDK<=8. There is no jre directory anymore.

JuergenFuessl:
This fix worked for us. The extension of the deprecated bootclasspath is not a good idea in Java 10. Although we also use Toolchains with Maven it seems to be working out of the box without manipulating the bootclasspath.

Issue #587, opened 2018-09-24
Build fails with latest jdt


cdietrich:
http://services.typefox.io/open-source/jenkins/job/xtext-xtend/job/master/708/console

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:1.2.0:compile (default-compile) on project org.eclipse.xtend.ide: Compilation failure: Compilation failure:
[ERROR] /var/jenkins_home/workspace/xtext-xtend_master-CZ6MQ7K3F54I6ECVAJOY5S5BWV6SPAS3CFPK73ZFQFHMBDDXMCYA/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/wizards/AbstractNewXtendElementWizardPage.java:[29]
[ERROR] import org.eclipse.jdt.internal.corext.codemanipulation.StubUtility;
[ERROR] ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[ERROR] The import org.eclipse.jdt.internal.corext.codemanipulation.StubUtility cannot be resolved
[ERROR] /var/jenkins_home/workspace/xtext-xtend_master-CZ6MQ7K3F54I6ECVAJOY5S5BWV6SPAS3CFPK73ZFQFHMBDDXMCYA/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/wizards/AbstractNewXtendElementWizardPage.java:[93]
[ERROR] return StubUtility.getCompilationUnitContent(compilationUnit, packageDeclaration, fileComment, typeComment, typeContent, lineSeparator);
```

cdietrich:
https://github.com/eclipse/eclipse.jdt.ui/commit/bc3bc858da1855f31cee9566912859faac05819e

cdietrich:
again we are hit by using internal api

cdietrich:
Fixed in 2.16

PR #596, opened 2018-10-07
[#587] made calls to StubUtility possible with jdt 4.9 and 4.10

cdietrich:
[#587] made calls to StubUtility possible with jdt 4.9 and 4.10

currently blocked by https://github.com/eclipse/xtext-xtend/pull/593


cdietrich:
@szarnekow @ArneDeutsch do you have any idea how to make this more elegantly?

szarnekow:
Unfortunately not from the top of my head. Resorting to reflection seems to be the only way in this case.

PR #599, opened 2018-10-08
[399591] Ensure the visibility of overridden method are properly applied

miklossy:
- Modify the XtendJvmModelInferrer to consider the visibility of the
overridden method when no explicit visibility modifier is defined.
- Implement corresponding Xtend Compiler test cases.

https://bugs.eclipse.org/bugs/show_bug.cgi?id=399591

### DIFF
--- a/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/CompilerBug399591Test.xtend
+++ b/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/CompilerBug399591Test.xtend
@@ -1,0 +1,98 @@
+/*******************************************************************************
+ * Copyright (c) 2018 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtend.core.tests.compiler
+
+import org.junit.Test
+
+/**
+ * @author miklossy - Initial contribution and API
+ */
+class CompilerBug399591Test extends AbstractXtendCompilerTest {
+
+	@Test def visibility_override_protected() {
+		'''
+			class C extends junit.framework.TestCase {
+				override setUp () {}
+			}
+		'''.assertCompilesTo('''
+			import junit.framework.TestCase;
+			
+			@SuppressWarnings("all")
+			public class C extends TestCase {
+			  protected void setUp() {
+			  }
+			}
+		''')
+	}
+
+	@Test def visibility_override_protected_protected() {
+		'''
+			class C extends junit.framework.TestCase {
+				protected override setUp () {}
+			}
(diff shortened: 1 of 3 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=399591

eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=399591

tivervac:
LGTM but I ought it best to wait for the official reviewer's review

cdietrich:
Has a lot of test failures

cdietrich:
http://services.typefox.io/open-source/jenkins/job/xtext-xtend/job/miklossy_master/lastCompletedBuild/

cdietrich:
java.lang.UnsupportedOperationException: TODO: import a functional handle on the type resolution that delegates to the best available (current, but evolving) result

miklossy:
The UnsupportedOperationException is thrown by the [DefaultReentrantTypeResolver](https://github.com/eclipse/xtext-extras/blob/master/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/internal/DefaultReentrantTypeResolver.java#L132-L134). I think the problem is that during resolving we try to access the resolved state that is currently being built. See also the comments from @kthoms on the [bugzilla ticket](https://bugs.eclipse.org/bugs/show_bug.cgi?id=399591):

> set the visibility to the overridden method's visibility, but I did not see if this could be safely computed in the inferrer. If the inferrer is not the right place to set the JvmOperation's visibility, where would it be?

cdietrich:
=> the override propsal and add unimplemented should specify the visibility explicitely too?

szarnekow:
> => the override propsal and add unimplemented should specify the visibility explicitely too?
if not public, yes.
I also added more thoughts to the review comment.

miklossy:
I would say we should postpone this issue to Release 2.17 to have more time to consider a proper solution.

szarnekow:
I'd like to have the validation in place for 2.16 so we can flag things already.

tivervac:
@szarnekow
1) I suppose this would be a temporary validation, until we can fix the root issue?
2) Do I understand it correctly that the current solution doesn't work because we're querying information from state that we're currently binding?
3) What's special about dispatch and create methods?

szarnekow:
> @szarnekow
> 
> 1. I suppose this would be a temporary validation, until we can fix the root issue?

Yes, that would be my preference. Validation and quickfix (if time permits) should already be a big step forward to mitigate the issue.

> 2. Do I understand it correctly that the current solution doesn't work because we're querying information from state that we're currently binding?

Exactly.

> 3. What's special about dispatch and create methods?

Dispatch and create methods do not have all parameter types available when the jvm model inferrer runs. They rely on resolved expressions. OTOH overload resolution does need to have signatures available if methods with the same name are called. That renders these cases more complex and special. They do make use of lazy JvmTypeReferences in their signatures.

cdietrich:
@szarnekow @miklossy how shall we continue

miklossy:
I would suggest to close this PR without merging it.

cdietrich:
will close the pr since it is incomplete

PR #618, opened 2018-10-22
[eclipse/xtext#1311] Make test independent from system encoding

kthoms:

### DIFF
--- a/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/richstring/RichStringEvaluationTest.java
+++ b/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/richstring/RichStringEvaluationTest.java
@@ -7,6 +7,7 @@
  *******************************************************************************/
 package org.eclipse.xtend.core.tests.richstring;
 
+import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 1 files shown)


org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/richstring/RichStringEvaluationTest.java
```diff
@@ -277,7 +278,7 @@ protected XtendFile file(String string, boolean validate) throws Exception {
 		XtextResourceSet set = resourceSetProvider.get();
 		String fileName = getFileName(string);
 		Resource resource = set.createResource(URI.createURI(fileName + ".xtend"));
-		resource.load(new StringInputStream(string), null);
+		resource.load(new LazyStringInputStream(string, StandardCharsets.ISO_8859_1), null);
```
szarnekow: I assume that we need to pass the encoding also with the loadOptions here.

szarnekow: I assume that we need to pass the encoding also with the loadOptions here.

PR #630, opened 2018-11-07
Keep the branchname in the TP but only replace the Jenkins URL

szarnekow:
Problem is, that the branchname is expected to come from a script parameter
but the sh script in the Jenkinsfile does not take any parameters. It always
replaces the stuff by 'master'

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,8 +45,8 @@ node {
 			targetfiles="$(find releng -type f -iname '*.target')"
 			for targetfile in $targetfiles
 			do
-				echo "Redirecting target platforms in $targetfile to $branchname"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -45,8 +45,8 @@ node {
 			targetfiles="$(find releng -type f -iname '*.target')"
 			for targetfile in $targetfiles
 			do
-				echo "Redirecting target platforms in $targetfile to $branchname"
-				sed_inplace "s?<repository location=\\".*/job/\\([^/]*\\)/job/[^/]*/?<repository location=\\"$JENKINS_URL/job/\\1/job/$escapedBranch/?" $targetfile
+				echo "Redirecting target platforms in $targetfile to $JENKINS_URL"
```
kthoms: `branchname`, `escaped` and `escapedBranch` should be removed also then.

PR #632, opened 2018-11-09
Make use of new JavaStringConverter

szarnekow:


szarnekow:
Note to self: Revert changes to target and pom files

PR #633, opened 2018-11-12
Clean up the org.eclipse.xtend.ide.tests/META-INF/MANIFEST.MF file.

miklossy:
- Eliminate duplicate entry 'org.eclipse.xtext.common.types.ui' from the
'Require-Bundle' section.


szarnekow:
Thank you

PR #640, opened 2018-11-22
[#582] Print warn message only when bootclasspath is not empty

kthoms:
Also fixed typo in the message.

### DIFF
--- a/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
+++ b/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
@@ -241,7 +241,9 @@ public class XtendBatchCompiler {
 	public void setBootClassPath(String bootClassPath) {
 		JavaVersion version = JavaVersion.fromQualifier(getJavaSourceVersion());
 		if (version.isAtLeast(JavaVersion.JAVA9)) {
-			log.warn("Option bootClassPath is only valid for Java 8 and lower. The velue '"+bootClassPath+"' will be ignored.");
+			if (!bootClassPath.isEmpty()) {
+				log.warn("Option bootClassPath is only valid for Java 8 and lower. The value '"+bootClassPath+"' will be ignored.");
+			}
 		} else {
 			this.bootClassPath = bootClassPath;
 		}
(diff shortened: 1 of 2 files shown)


kthoms:
@cdietrich I did another small change, please check again.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
```diff
@@ -241,7 +241,9 @@ public void setClassPath(String classPath) {
 	public void setBootClassPath(String bootClassPath) {
 		JavaVersion version = JavaVersion.fromQualifier(getJavaSourceVersion());
 		if (version.isAtLeast(JavaVersion.JAVA9)) {
-			log.warn("Option bootClassPath is only valid for Java 8 and lower. The velue '"+bootClassPath+"' will be ignored.");
+			if (!bootClassPath.isEmpty()) {
```
cdietrich: should be check for nullOrEmpty?

org.eclipse.xtend.maven.plugin/src/main/java/org/eclipse/xtend/maven/AbstractXtendCompilerMojo.java
```diff
@@ -179,17 +179,18 @@ protected void compile(String classPath, List<String> sourcePaths, String output
...
-			if (version.isAtLeast(JavaVersion.JAVA9)) {
-				return ""; // bootclasspath only supported on Java8 and older
-			}
-		}
 		Toolchain toolchain = toolchainManager.getToolchainFromBuildContext("jdk", session);
 		if (toolchain instanceof DefaultJavaToolChain) {
 			DefaultJavaToolChain javaToolChain = (DefaultJavaToolChain) toolchain;
 			getLog().info("Using toolchain " + javaToolChain);
 
+			if (javaSourceVersion != null) {
```
cdietrich: what is with the else case?

kthoms: When no javaSourceVersion is set it cannot be decided that this is run on Java9+.
And actually it won't be null since it defaults to `maven.compiler.source`. We could argue that the null guard might not be necessary here.

cdietrich:
:+1: LGTM. lets see what ci says

Issue #639, opened 2018-11-20
Fix Build Issues with Scheduling rules


cdietrich:
```
Fehlermeldung

Attempted to beginRule: MultiRule[P/first.p384008,P/.org.eclipse.jdt.core.external.folders], does not match outer scope rule: MultiRule[P/first.p384008,P/second.p384008,P/test.project] - See error log for details, The type A is already defined in B.xtend., Attempted to beginRule: MultiRule[P/first.p384008,P/.org.eclipse.jdt.core.external.folders], does not match outer scope rule: MultiRule[P/first.p384008,P/second.p384008,P/test.project] - See error log for details expected:<1> but was:<3>

Stacktrace

java.lang.AssertionError: Attempted to beginRule: MultiRule[P/first.p384008,P/.org.eclipse.jdt.core.external.folders], does not match outer scope rule: MultiRule[P/first.p384008,P/second.p384008,P/test.project] - See error log for details, The type A is already defined in B.xtend., Attempted to beginRule: MultiRule[P/first.p384008,P/.org.eclipse.jdt.core.external.folders], does not match outer scope rule: MultiRule[P/first.p384008,P/second.p384008,P/test.project] - See error log for details expected:<1> but was:<3>
	at org.eclipse.xtend.ide.tests.validation.UniqueClassNameValidatorUITest.testTwoXtendFilesSameProject(UniqueClassNameValidatorUITest.java:65)

```

cdietrich:
c @kthoms

cdietrich:
can be reproduce locally when adding

`-pluginCustomization /data/xtext-master/git/xtext-xtend/releng/org.eclipse.xtend.tycho.tests.parent/pluginCustomization.ini` to launch config

cdietrich:
```
	@Override
	public ISchedulingRule getRule(int kind, Map<String, String> args) {

		switch (preferences.schedulingOption) {
			case NULL:
				return null;
			case WORKSPACE:
				return getProject().getWorkspace().getRoot();
			case PROJECT:
				return getProject();
			case ALL_XTEXT_PROJECTS: {
				IProject implicitProject = getProject().getWorkspace().getRoot().getProject(".org.eclipse.jdt.core.external.folders");
				Stream<IProject> implicit = Stream.of();
				if (implicitProject != null) {
					implicit = Stream.of(implicitProject);
				}
[...]

PR #641, opened 2018-11-23
[#639] wrap build into workspace operations

cdietrich:
[#639] wrap build into workspace operations


cdietrich:
i still dont know if this is the correct solution or if we should change the builder to include the external folder project if available. but if so wont that hinder us running in parallel with java project no related to Xtext.

kthoms:
Right, it would. Like for the JavaBuilder, too.
The problem is the potential classpath change. Can't we forbid this during a build?

cdietrich:
no its a feature that Xtext?/Xbase?/Xtend magically make the gen folder a source folder as far as i have understood. i dont know if there are other similar things that change the classpath somewhere else burried in our code

Issue #662, opened 2018-12-10
ContentAssistInElseBlockTest fails if first test


cdietrich:
ContentAssistInElseBlockTest fails if first test.
maybe we need to make sure e.g. TargetPlatformUtil is called on it. (or something else for proper init)
(am not sure if other tests are affected too)

PR #663, opened 2018-12-10
[#662] fixed failing test

cdietrich:
[#662] fixed failing test


cdietrich:
its always necessary. at least if we dont want to be order dependent

PR #666, opened 2018-12-11
bootstraping version to 2.16.0

pngowda:
Signed-off-by: Prajwal Gowda <pgowda@itemis.de>
@cdietrich  could you please review it


cdietrich:
thank you @pngowda

PR #667, opened 2018-12-11
eclipse/xtext-eclipse#920 Check in xtend-gen for all examples.

ArneDeutsch:


ArneDeutsch:
Verified that the created JAR files contain the xtend-gen folder and all java files are there.

Issue #669, opened 2018-12-12
Clean and Archive screenshots


cdietrich:
- clean build should clean the screenshot dir for swtbot
- screenshots in the screenshot dir should be archived by the build

ArneDeutsch:
Done.

cdietrich:
thanks @ArneDeutsch

PR #670, opened 2018-12-12
#669 Archive screenshots and clean them before build.

ArneDeutsch:


cdietrich:
(did not check if actually works but looks good to me)

PR #673, opened 2018-12-17
[eclipse/xtext#1363] prepare build against 2019-03

cdietrich:
[eclipse/xtext#1363] prepare build against 2019-03


cdietrich:
cannot be merged before 4.10/201812 is released.
change for the eclipse update site will be done as soon as /2019-03 is available

PR #679, opened 2019-01-03
Xtext eclipse 813 automated ui tests

ArneDeutsch:


cdietrich:
what is the intention behind this changes?

ArneDeutsch:
There are some (very few) existing UI tests for Xtend. I have moved the infrastructore to a common subproject to avoid the redundance.

cdietrich:
but why is stuff red then

ArneDeutsch:
Damn. Multip-Repo-Issue. The shared project in moved to umbrella ... not availanble in Xtend ... argh ...

ArneDeutsch:
Ok, made up my mind. Probably best to accept a bit of duplication here. Will keep the infrastructure as required by this test inside this project. Actually the duplication is not that big because the parts of the UI accessed by this test are other parts then from the test in umbrella. Will close this request.

Issue #671, opened 2018-12-12
Xtend IDE: Implement Quickfix to the 'unused private member' validation issue


miklossy:
The Xtend IDE should provide a Quickfix that removes the `unused private member`:
![screenshot](https://user-images.githubusercontent.com/20393472/49875934-8811a100-fe22-11e8-9f5c-6bc88dbea5ea.png)

szarnekow:
Also for local vars

miklossy:
Fixed in 2.17.

PR #689, opened 2019-01-16
[#671] Add Quickfix to the 'unused private member' validation issue.

miklossy:
- Modify the XtendQuickfixProvider to ensure that the unused private
members can be removed by quickfixes.
- Implement corresponding QuickfixTest test cases.


org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/quickfix/XtendQuickfixProvider.java
```diff
@@ -690,6 +691,20 @@ public void changeJUnitMethodReturnTypeToVoid(final Issue issue, IssueResolution
 		});
 	}
 	
+	@Fix(IssueCodes.UNUSED_PRIVATE_MEMBER)
+	public void removeUnusedPrivateMember(final Issue issue, IssueResolutionAcceptor acceptor) {
+		// use the same label, description and image
+		// to be able to use the quickfixes (issue resolution) in batch mode
+		String label = "Remove.";
+		String description = "Remove the unused private member.";
+		String image = "fix_indent.gif";
```
holgerschill: We will use the remove icon from jdt instead.

miklossy: Thanks for the feedback, it is a good idea to use the same symbols as jdt uses. I updated the PR as requested.

PR #700, opened 2019-01-30
[eclipse/xtext#1283] update to MWE 2.10 M2

cdietrich:
[eclipse/xtext#1283] update to MWE 2.10 M2


releng/org.eclipse.xtend.target/org.eclipse.xtend.target-photon.target
```diff
@@ -113,7 +113,7 @@
 	<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="false" type="InstallableUnit">
 		<unit id="org.eclipse.emf.mwe2.lib" version="0.0.0"/>
 		<unit id="org.eclipse.emf.mwe2.launcher.feature.group" version="0.0.0"/>
-		<repository location="http://download.eclipse.org/modeling/emft/mwe/updates/releases/2.9.1/"/>
+		<repository location="http://download.eclipse.org/modeling/emft/mwe/updates/milestones/S201901281525/"/>
```
kthoms: Wondering if SimRel release targets might want to consume the MWE version that was on the release repository, i.e. don't include the MWE update site at all except for in 'latest' and move the units to the release repository. WDYT?

cdietrich: did not get that. can you please rephrase

kthoms: MWE bundles are also in the release repository, here http://download.eclipse.org/releases/photon
It is version 2.9.1 there, but shouldn't the "Photon" target offer what the Photon platform offers?

cdietrich: hmmm dont know. @szarnekow what do you think?

cdietrich: would also mean a change for oxygen / 2.7.1

szarnekow: I think this would imply unnecessary maintenance overhead.

kthoms: Actually we are doing now maintenance change and would do another when the MWE release repo is available. And next time a new MWE repo URL comes up again, and again...
So from my point it would actually reduce the maintenance effort.

cdietrich: ok @kthoms could you fix both prs and the oxygen targets as well?

szarnekow: I'm not sure about this. Coupling to an older mwe for older TPs (that is how I understood your proposal) would increase the effort.

cdietrich: Karstens assumption is that these won’t change avd only latest needs adoption

kthoms: >could you fix both prs and the oxygen targets as well?

Will do

releng/org.eclipse.xtend.target/org.eclipse.xtend.target-oxygen.target
```diff
@@ -118,6 +113,9 @@
 	
 	<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
 		<unit id="org.junit" version="0.0.0"/>
+		<unit id="com.google.inject" version="0.0.0"/>
```
cdietrich: what is behind this change?

kthoms: Good catch, should not be there. Copy/pasted it from some other target file.

kthoms: Actually this was present in the other targets in the Orbit section. Doesn't hurt to have it here explicity.

PR #735, opened 2019-02-26
[xtext-xtend#734] Refined approach to registry setup and restore in tests

szarnekow:


szarnekow:
This PR has the cherry-picked fixes from M3, too.

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/RuntimeInjectorProvider.java
```diff
@@ -28,29 +28,32 @@
...
+
 	protected Injector internalCreateInjector() {
-	    return new RuntimeTestSetup().createInjectorAndDoEMFRegistration();
+		return new RuntimeTestSetup().createInjectorAndDoEMFRegistration();
 	}
 
 	@Override
 	public void restoreRegistry() {
 		stateBeforeInjectorCreation.restoreGlobalState();
+		stateBeforeInjectorCreation = null;
```
tivervac: What exactly does moving the `stateBeforeInjectorCreation` do? We've had it in the same place for quite some time and don't appear to suffer from flakyness

szarnekow:
The problem becomes apparent with a sequence of RuntimeInjectorProvider, XtendIDEInjectorProvider, RuntimeInjectorProvider and again XtendIDEInjectorProvider. In the forth test class, the validator registries are in a corrupted state, eg no XtendValidator is available anymore. I plan to fix this for all injector providers and provide a test.

szarnekow:
The approach in this PR does not work, I guess. A lazily created injector of another language (eg via an executable extension factory) will populate the registry and the settings will be reverted after the test but never put into the registry again.

The change to the IDE injector provider is likely unnecessary.

szarnekow:
Build is green here https://ci.eclipse.org/xtext/job/xtext-xtend/job/sz_fixInjectorProvider/3/
and here http://services.typefox.io/open-source/jenkins/job/xtext-xtend/job/sz_fixInjectorProvider/

Issue #736, opened 2019-02-28
Xtend Quickfix should also remove the unused imports and all writing accesses to the field


miklossy:
Given is the following Xtend file:

```
package foo

import java.util.Date
import java.util.List

class A {
  List<Date> a
  
  new() {
    a = newArrayList
  }
}
```
Currently, the Xtend Quickfix Remove unused member only removes the declaration of an unused field
![screencast](https://user-images.githubusercontent.com/20393472/53564454-42c65b00-3b57-11e9-97ba-d9afda5450d3.gif)

The Quickfix should clean up as much as possible, such as:

- the import statements that become unused after removing the unused field
- all writing accesses to the field

@loradd  Would you like to take care of this issue? Would be awesome...

miklossy:
**Hint:** The JDT removes 'a', but keeps the assignments with side-effects.

loradd:
Ok, I'll take care of it!

cdietrich:
@miklossy is this fixed now?

miklossy:
Thanks for your contribution! I merged yor PR into the master branch! This issue is fixed in 2.18.

PR #740, opened 2019-03-05
[issue#736] Xtend Quickfix should also remove the unused imports and all writing accesses to the field

loradd:
Xtend Quickfix should also remove the unused imports and all writing accesses to the field #736

Signed-off-by: Lorenzo Addazi <addazi.lorenzo@gmail.com>

### DIFF
--- a/org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/QuickfixTest.xtend
+++ b/org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/QuickfixTest.xtend
@@ -4182,14 +4182,23 @@ class QuickfixTest extends AbstractXtendUITestCase {
 	@Test
 	def void fixUnusedPrivateField() {
 		create('Foo.xtend', '''
+			import java.util.List
+			import java.util.Date
+			
 			class Foo {
-				val a| = 42
+				List<Date> |a
+				new () {
+					a = newArrayList
+				}
 			}
 		''')
 		.assertIssueCodes(UNUSED_PRIVATE_MEMBER)
 		.assertResolutionLabels("Remove member.")
 		.assertModelAfterQuickfix('''
 			class Foo {
+			
+				new () {
+				}
 			}
 		''')
 	}
--- a/org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/QuickfixTest.xtend
+++ b/org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/QuickfixTest.xtend
@@ -4182,6 +4182,22 @@ class QuickfixTest extends AbstractXtendUITestCase {
 	@Test
 	def void fixUnusedPrivateField() {
 		create('Foo.xtend', '''
+			class Foo {
+				val a| = 42
+			}
+		''')
+		.assertIssueCodes(UNUSED_PRIVATE_MEMBER)
+		.assertResolutionLabels("Remove member.")
+		.assertModelAfterQuickfix('''
+			class Foo {
+				
+			}
+		''')
+	}
+	
+	@Test
+	def void fixUnusedPrivateFieldWithAssignmentsAndImports() {
+		create('Foo.xtend', '''
 			import java.util.List
 			import java.util.Date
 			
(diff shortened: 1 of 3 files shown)


org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/quickfix/XtendQuickfixProvider.java
```diff
@@ -741,10 +745,25 @@ public void removeUnusedPrivateMember(final Issue issue, IssueResolutionAcceptor
 		String label = "Remove member.";
 		String description = "Remove the unused private member.";
 		String image = "delete_edit.png";
-		acceptor.acceptMulti(issue, label, description, image, (ICompositeModification<XtendMember>) (element, ctx) -> {
-			ctx.setUpdateCrossReferences(false);
-			ctx.setUpdateRelatedFiles(false);
-			ctx.addModification(element, ele -> EcoreUtil.remove(ele));
+		acceptor.accept(issue, label, description, image, new ISemanticModification() {
```
kthoms: why does this not use acceptMulti? Does the quickfix work on multiple issues?

szarnekow: Since this is a rather complicated quickfix (preserving side effect semantics et al), I'd like to keep this to a bare minimum for now.

org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/QuickfixTest.xtend
```diff
@@ -4182,14 +4182,23 @@ class QuickfixTest extends AbstractXtendUITestCase {
...
+					a = newArrayList
+				}
 			}
 		''')
 		.assertIssueCodes(UNUSED_PRIVATE_MEMBER)
 		.assertResolutionLabels("Remove member.")
 		.assertModelAfterQuickfix('''
 			class Foo {
+			
+				new () {
```
szarnekow: We should try to keep the side effects on the right hand side of assignments. newArrayList is annotated as `@Pure` so it is valid to be removed. But lets say the code was `s = toString()` - in that case we'd need to keep the invocation of `toString()`

org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/quickfix/XtendQuickfixProvider.java
```diff
@@ -741,10 +745,25 @@ public void removeUnusedPrivateMember(final Issue issue, IssueResolutionAcceptor
...
+			@Override
+			public void apply(EObject element, IModificationContext context) throws Exception {
+				if (element instanceof XtendField) {
+					final JvmField field = associations.getJvmField((XtendField)element);
+					final IXtextDocument document = context.getXtextDocument();
+					document.modify(new IUnitOfWork.Void<XtextResource>() {
+						@Override
+						public void process(XtextResource resource) throws Exception {
+							EcoreUtil.remove(element);
+							EcoreUtil.removeAll(EcoreUtil2.eAllContentsAsList(resource).stream()
```
szarnekow: Since we'd likely want to keep a bunch of the side effects, a text based quick fix may be the better choice in the long run.

org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/quickfix/XtendQuickfixProvider.java
loradd: I was not sure how to remove all unused imports without reimplementing `OrganizeImportsHandler#doOrganizeImports`, which requires an `IXtextDocument`. Has this been reimplemented somewhere else?

@szarnekow : do you mean not removing assignments and unused imports?

szarnekow: No, I was referring to the suggestion of using acceptMulti.

loradd:
How can I trigger another build?

cdietrich:
did trigger it for you

loradd:
Thank you!

kthoms:
> How can I trigger another build?

You could trigger it by pushing again. If there is no real change, just use amend with empty commit. Or even better: rebase the branch against master.

cdietrich:
@miklossy what is the state of review

loradd:
@szarnekow it seems that the quickfix is entirely undone in a single step, and this is the expected behaviour, isn't it? How could I include it in the unit tests?

szarnekow:
> How could I include it in the unit tests?

I'm not sure. I think this is a case of 'manual testing is sufficient for now'. Thanks for confirming the behavior.

PR #745, opened 2019-03-07
 [eclipse#395] Try with resources 

epoell:
Validate, that a given resources implements the AutoClosable Interface
and that mark as an error otherwise.

Also support content proposals for defined resources.

Finally check the whole functionality with JUnit Tests.
Both, the Error marking, as well as the compilation to correct Java Code is
checked for Java 8 and for Java versions, that do not support try with
resources.


org.eclipse.xtend.core.tests.java8/src/org/eclipse/xtend/core/tests/java8/compiler/TryWithResourcesTestJava8.xtend
```diff
@@ -0,0 +1,245 @@
...
+	@Test
+	def void test_normalTryWithoutResource() {
+		'''val br = new BufferedReader(new StringReader(s));
+		try 
+			br.readLine()
+		finally
+			if(br !== null)
+				br.close()'''.test2('''      StringReader _stringReader = new StringReader(this.s);
+      final BufferedReader br = new BufferedReader(_stringReader);
+      try {
```
kthoms: indentation looks weird. are spaces / tabs mixed? Please always turn on "Show Whitespace" for the text editor in preferences.

epoell: Indeed the indentation is mixed, as the result of AbstractXtendCompilerTest uses spaces in the resulting generated JavaCode instead of tabs and I use a plain String comparision for asserting the right result (which lets the test fail for tabs).

Shall I thus use spaces instead of tabs thourough the whole class to make it uniform?

szarnekow: Having spaces within the string literals should be fine. Usually multi-line strings are formatted with a leading line wrap and indentation:
```
  '''
     some content
  '''.doStuff()
```

org.eclipse.xtend.core.tests.java8/src/org/eclipse/xtend/core/tests/java8/compiler/TryWithResourcesTestJava8.xtend
```diff
@@ -0,0 +1,245 @@
...
+		}'''.test0('''      try (final AutoCloseable someCloseable = ((AutoCloseable) () -> {
+        System.out.println("Closing");
+      })) {
+      }
+	  ''')
+	}
+	
+	@Test
+	def void test_twoResources() {
+		'''try (val sr = new StringReader(s); val buffy = new BufferedReader(sr)) {
```
szarnekow: Could you please add a case where the constructor of the resource declared an IOException, e.g. FileReader?

epoell: Do you mean like `try (val a = new FileReader(new File("\\home/docs/text.txt")))`?
The file is not existent and even not a legal path, thus an IOException should be thrown. However, this is not thrown until the JavaCode is compiled, isn't it?

And if the case above is what you had in mind, is the following code according to your expectation?
```
   try {
      int _xtrycatchfinallyexpression = (int) 0;
      try (final FileReader a = new Function0<FileReader>() {
        public FileReader apply() {
          try {
            File _file = new File("\\home/docs/text.txt");
            return new FileReader(_file);
          } catch (Throwable _e) {
            throw Exceptions.sneakyThrow(_e);
          }
        }
      }.apply()) {
        int _read = a.read();
        _xtrycatchfinallyexpression = _read;
      }
      return _xtrycatchfinallyexpression;
    } catch (Throwable _e) {
      throw Exceptions.sneakyThrow(_e);
    }
```

szarnekow: Exactly what I had in mind

epoell:
Hello
My commit cannot be build with the error message:
"Execution failed for task ':org.eclipse.xtend.core:generateXtext'.
> Could not resolve all files for configuration ':org.eclipse.xtend.core:compileClasspath'.
   > Could not find org.eclipse.xtext:org.eclipse.xtext.xbase:2.18.0-SNAPSHOT.
     Required by:
         project :org.eclipse.xtend.core"

What should I do, to correct this error?
Is this due to my possibly faulty MANIFEST.MF files (see [the other discussion](https://github.com/eclipse/xtext-extras/pull/390#pullrequestreview-212277400)) How can I rebuild them?

cdietrich:
this pr is replaced by https://github.com/eclipse/xtext-xtend/pull/769. thx @epoell

PR #771, opened 2019-04-03
[eclipse/xtext#1309] Provide Jenkinsfile for Eclipse CBI

kthoms:
- externalized build steps into shell scripts for convenient local usage
- added Slack notifications
- configure Maven default goal
- Make RichStringEvaluationTest test independent from system encoding
- Ignore failing tests from AbstractReusableActiveAnnotationTests
- Fix encoding issues


maven-pom.xml
```diff
@@ -20,8 +20,9 @@
 	</modules>
 
 	<build>
+		<defaultGoal>deploy</defaultGoal>
```
szarnekow: other jobs used install?

tycho-pom.xml
```diff
@@ -31,5 +31,9 @@
 		<module>org.eclipse.xtend.doc</module>
 		<module>org.eclipse.xtend.examples</module>
 	</modules>
+	
+	<build>
+		<defaultGoal>verify</defaultGoal>
```
szarnekow: deploy vs install vs verify?

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/macro/AbstractReusableActiveAnnotationTests.xtend
```diff
@@ -2922,6 +2922,7 @@ abstract class AbstractReusableActiveAnnotationTests {
 		]
 	}
 	
+	@Ignore("FIXME: Issue 1311 - Make system encoding independent")
```
cdietrich: ?

tycho-pom.xml
kthoms: for the maven plugins deploy is unfortunately necessary for now. they deploy to a local file repo, which is not nice. but different story.

maven-pom.xml
kthoms: for the maven plugins deploy is unfortunately necessary for now. they deploy to a local file repo, which is not nice. but different story.

tycho-pom.xml
kthoms: OK, here install would be better.

Issue #617, opened 2018-10-22
AbstractReusableActiveAnnotationTests - Make independent from system encoding


kthoms:
The following methods fail when the test is executed with US-ASCII:
- testFileSystemSupport_01
- testImportFromTypeReference_01
- testImportFromTypeReference_02

Subtask of eclipse/xtext#1311

kthoms:
For reproduction execute the test with `-Dfile.encoding=US-ASCII`

kthoms:
Done

PR #772, opened 2019-04-04
[#617] Make independent from system encoding

kthoms:


kthoms:
Does not work yet

PR #776, opened 2019-04-05
Change defaultGoal to install

kthoms:
install is required to trigger the maven-antrun-plugin in
org.eclipse.xtend.p2-repository, which copies the resulting p2
repository to the target build folder. Without, the p2 repository is not
archived.


cdietrich:
please dont merge before we have verified this

Issue #777, opened 2019-04-06
JavaClasspathTest fails on "latest"


cdietrich:
JavaClasspathTest fails on "latest"

```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertFalse(Assert.java:64)
	at org.junit.Assert.assertFalse(Assert.java:74)
	at org.eclipse.xtend.ide.tests.buildpath.JavaClasspathTest$1.run(JavaClasspathTest.java:77)
	at org.eclipse.xtext.testing.logging.LoggingTester.captureLogging(LoggingTester.java:353)
	at org.eclipse.xtend.ide.tests.buildpath.JavaClasspathTest.testNoJavaInClasspath(JavaClasspathTest.java:46)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
[...]

PR #781, opened 2019-04-15
[#777] Make JavaClasspathTest compatible to latest

szarnekow:
[#777] Make JavaClasspathTest compatible to latest


cdietrich:
@szarnekow does this mean we dont have to change anything function wise?

szarnekow:
As far as I can tell, the recovered bindings are exactly what we wanted to have right from the beginning. The logged errors were more of an annoyance than a feature.

PR #790, opened 2019-04-29
[eclipse/xtext-extras#415] Tests for compiling try-with-resource

max-leuthaeuser:
Adapted tests for improved exception handling for try-with-resource.
For: https://github.com/eclipse/xtext-extras/pull/416

**Note:** won't run through CI as this https://github.com/eclipse/xtext-extras/pull/416 is not merged yet.

Signed-off-by: Max Leuthäuser <max.leuthaeuser@itemis.de>


org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/TryWithResourcesTest.xtend
```diff
@@ -75,21 +75,63 @@ class TryWithResourcesTest extends AbstractXtendCompilerTest {
...
+			    } finally {
+			      List<Throwable> _ts = new ArrayList<Throwable>();
+			      if (a != null) {
+			        try {
+			          a.close();
+			        } catch (Throwable _t_1) {
+			          _ts.add(_t_1);
+			        }
+			      }
+			      if(!_ts.isEmpty()) throw _ts.get(0);
```
szarnekow: I wonder if this is the right exception to be propagated. According to https://www.javaspecialists.eu/archive/Issue259.html, it should be the throwable from the try block if there is any.

max-leuthaeuser: Hm... maybe I misunderstood something there but the exception from the try is always thrown correctly. With my implementation the Throwables that might be thrown during close are stored additionally and propagated later on. The article you have linked is using _addSuppressed_ while I am using an additional list.

szarnekow: To me it looks like you are throwing from a finally block. The thrown exception will likely cause the original exception to be lost

max-leuthaeuser: Yes, you are right. Then I will have to use the "store the exception from the try block and use addSupressed within the finally" pattern.

szarnekow: I don’t know if addSuppressed was available before Java7 so the list was a good idea

max-leuthaeuser: Unfortunately, it's only available since 1.7 according to the JavaDoc.

szarnekow:
Thank you

PR #793, opened 2019-05-03
[eclipse/xtext-core#1137] Filter static methods in Delegate active annotation

cdietrich:
[eclipse/xtext-core#1137] Filter static methods in Delegate active annotation


cdietrich:
impl at https://github.com/eclipse/xtext-lib/pull/184

Issue #800, opened 2019-05-07
Failing Tests with Java 11


cdietrich:
- org.eclipse.xtend.core.tests.typesystem.BoundTypeArgumentMergerTest
- org.eclipse.xtend.core.tests.performance.PerformanceTest
- org.eclipse.xtend.core.tests.typesystem.CommonSuperTypeTest
- org.eclipse.xtend.core.tests.compiler.CompilerBug457539Test
- org.eclipse.xtend.core.tests.compiler.ConfiguredCompilerTest
- org.eclipse.xtend.core.tests.compiler.batch.TestBatchCompiler
- org.eclipse.xtend.core.tests.compiler.CompilerBug406425Test
- org.eclipse.xtend.core.tests.javaconverter.JavaConverterTest
- org.eclipse.xtend.core.tests.imports.ImportsCollectorTests
- org.eclipse.xtend.core.tests.java8.compiler.CompilerBug457539Test
- org.eclipse.xtend.core.tests.java8.javaconverter.Java8ConverterTest

cdietrich:
create https://github.com/eclipse/xtext-xtend/issues/804 for urlclassloader problem

cdietrich:
Fixed normal tests. now looking at longrunning tests

cdietrich:
Eclipse

u.a.

```
ERROR (org.eclipse.xtext.diagnostics.Diagnostic.Syntax) 'no viable alternative at input '['' on XMemberFeatureCall, offset 45, length 1
ERROR (org.eclipse.xtext.diagnostics.Diagnostic.Syntax) 'extraneous input ']' expecting '}'' on XBlockExpression, offset 46, length 1
ERROR (org.eclipse.xtext.xbase.validation.IssueCodes.invalid_number_of_arguments) 'Invalid number of arguments. The method accept(I) is not applicable without arguments' on XMemberFeatureCall, offset 39, length 6
WARNING (org.eclipse.xtend.core.validation.IssueCodes.wrong_file) 'The type name 'D' doesn't match the file name 'Dummy'' on XtendClass, offset 6, length 1
ERROR (org.eclipse.xtend.core.validation.IssueCodes.xbase_lib_not_on_classpath) 'Couldn't find the mandatory library 'org.eclipse.xtext.xbase.lib' 2.8.0 or higher on the project's classpath.' on XtendFile, offset 0, length 51
[...]

PR #803, opened 2019-05-10
[#800] fixed tests on java 11

cdietrich:


org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/batch/TestBatchCompiler.xtend
```diff
@@ -495,6 +497,8 @@ class TestBatchCompiler {
 	
 	@Test
 	def void testGeneratedAnnotation() {
+		if (AbstractXtendTestCase.isJava11OrLater)
```
szarnekow: Cosmetics: Could be `Assume.assumeTrue(isJava11OrLater)`

cdietrich: fixed

PR #808, opened 2019-05-13
fixed determineJava11OrLater

cdietrich:


org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/AbstractXtendTestCase.java
```diff
@@ -62,8 +62,13 @@ public static boolean isJava11OrLater() {
 	private static boolean determineJava11OrLater() {
 		String javaVersion = System.getProperty("java.version");
 		try {
-			int version = Integer.parseInt(javaVersion);
-			return version >= 11;
+			Pattern p = Pattern.compile("(\\d+)(.)*");
```
szarnekow: Alternatively we could check CharSequence.class.getMethod(..) and catch NoSuchMethod.

PR #817, opened 2019-05-17
Use classgraph library to scan classpath contents

szarnekow:


org.eclipse.xtend.core.tests/.classpath
```diff
@@ -14,6 +14,7 @@
 	</classpathentry>
 	<classpathentry kind="src" output="bin/test" path="testdata">
 		<attributes>
+			<attribute name="ignore_optional_problems" value="true"/>
```
cdietrich: @szarnekow is this a thing gradle adds or do we miss a config in gradle?

szarnekow: I changed this in the UI. My bad.

cdietrich: do you think we should add this to gradle for all (tests?) projects?

szarnekow: It would make sense for the source folders, that do contain test-java-code

Issue #818, opened 2019-05-17
Ide Tests fail cause of ClassGraph


cdietrich:
```
!ENTRY org.apache.log4j 4 0 2019-05-17 20:05:18.477
!MESSAGE org.eclipse.xtend.ide.internal.XtendActivator  - Failed to create injector for org.eclipse.xtend.core.Xtend

10   [main] ERROR org.eclipse.xtend.ide.internal.XtendActivator  - com.google.inject.internal.util.$ComputationException: java.lang.NoClassDefFoundError: io/github/classgraph/ClassInfo
com.google.inject.internal.util.$ComputationException: com.google.inject.internal.util.$ComputationException: java.lang.NoClassDefFoundError: io/github/classgraph/ClassInfo
	at com.google.inject.internal.util.$MapMaker$StrategyImpl.compute(MapMaker.java:553)
	at com.google.inject.internal.util.$MapMaker$StrategyImpl.compute(MapMaker.java:419)
	at com.google.inject.internal.util.$CustomConcurrentHashMap$ComputingImpl.get(CustomConcurrentHashMap.java:2041)
	at com.google.inject.internal.FailableCache.get(FailableCache.java:50)
	at com.google.inject.internal.ConstructorInjectorStore.get(ConstructorInjectorStore.java:49)
	at com.google.inject.internal.ConstructorBindingImpl.initialize(ConstructorBindingImpl.java:125)
	at com.google.inject.internal.InjectorImpl.initializeJitBinding(InjectorImpl.java:521)
	at com.google.inject.internal.InjectorImpl.createJustInTimeBinding(InjectorImpl.java:847)
	at com.google.inject.internal.InjectorImpl.createJustInTimeBindingRecursive(InjectorImpl.java:772)
	at com.google.inject.internal.InjectorImpl.getJustInTimeBinding(InjectorImpl.java:256)
	at com.google.inject.internal.InjectorImpl.getBindingOrThrow(InjectorImpl.java:205)
	at com.google.inject.internal.InjectorImpl.getInternalFactory(InjectorImpl.java:853)
	at com.google.inject.internal.FactoryProxy.notify(FactoryProxy.java:46)
	at com.google.inject.internal.ProcessedBindingData.runCreationListeners(ProcessedBindingData.java:50)
	at com.google.inject.internal.InternalInjectorCreator.initializeStatically(InternalInjectorCreator.java:133)
[...]

PR #819, opened 2019-05-17
[#818] better decoupling from standalone and eclipse astparserfactory

cdietrich:
[#818] better decoupling from standalone and eclipse astparserfactory

### DIFF
--- a/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/javaconverter/EclipseASTParserFactory.xtend
+++ b/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/javaconverter/EclipseASTParserFactory.xtend
@@ -34,7 +34,7 @@ class EclipseASTParserFactory extends org.eclipse.xtend.core.javaconverter.ASTPa
 			}
 			return new ASTParserWrapper(projlevel, parser)
 		}
-		return super.createJavaParser(context)
+		throw new IllegalStateException("did not find java project")
 	}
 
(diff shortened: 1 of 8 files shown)


org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/javaconverter/EclipseASTParserFactory.xtend
```diff
@@ -34,7 +34,7 @@ class EclipseASTParserFactory extends org.eclipse.xtend.core.javaconverter.ASTPa
 			}
 			return new ASTParserWrapper(projlevel, parser)
 		}
-		return super.createJavaParser(context)
+		throw new IllegalStateException("did not find java project")
```
cdietrich: i wonder which usecase this addressed and what we should do here.
if there is no java project does it make sense to use the classpath of xtend.ide?
(beside adding clasgraph to orbit)

cdietrich:
Looks like that does not completely help
https://ci.eclipse.org/xtext/job/xtext-xtend/job/cd_issue818/2/console

cdietrich:
looks like there are ide tests that use runtime injector
org.eclipse.xtend.ide.tests.AbstractXtendTestCase

cdietrich:
=> Options are:
- find solution
- revert changes and find clean solution
- add classgraph to orbit (might be needed for @ArneDeutsch anyway)

szarnekow:
I’ll file the CQ for orbit

cdietrich:
Obsolete

PR #820, opened 2019-05-18
[#818] Avoid the ClasspathScanner on signature of AstParserFactory

szarnekow:


cdietrich:
I still wonder what happens if the eclipse impl calls super here and if we should avoid calling super there

szarnekow:
This is supposed to be an intermediate solution until we can consume ClassGraph from orbit.

cdietrich:
Tests look good

PR #860, opened 2019-06-25
[eclipse/xtext#1431] Tweak resource constraints

kthoms:
Tests have shown that at most 2 executors will usually be available at
most and these resource configurations seem to make optimal usage of
available resources.
- Container 'jnlp' is working with 0.4Gi memory (default is 0.5Gi) and
0.2 cpu (this is the default, lower value leads to slower agent
provisioning)
- Build container can use at max 3.6Gi memory and 1.0 cpu. Higher values
will lead to only 1 executor.


kthoms:
Sorry, accidently merged. Shall we revert?

cdietrich:
no looks ok

PR #867, opened 2019-07-02
Parallelize stages 'Maven Plugin Build' & 'Maven Tycho Build'

kthoms:
This change enables parallelization of the two Maven stages. It looks as this saves a few minutes in the build and looks stable.

without parallel execution (master)
screenshot 128

with parallel execution
screenshot 129

### DIFF
--- a/CBI.Jenkinsfile
+++ b/CBI.Jenkinsfile
@@ -121,7 +121,7 @@ spec:
     }
 
     stage('Maven Build & Test') {
-      stages { // TODO use of parallel { here kills Tycho process with OOM
+      parallel {
         stage('Maven Plugin Build') {
           steps {
             sh """


cdietrich:
i still dislike this for the broken progress reporting
am not sure if we do something wrong or jenkins is really that bad

Hejado:
Did the parallel stage spawn a second agent/executor? If so, could we test if we're safe against blocking by duplicating this branch and triggering two simultaneous builds?

If both parallel stages actually run inside the same agent and don't block a second one, this would be approvable imho.

kthoms:
> Did the parallel stage spawn a second agent/executor? 

No, it ran in the same executor.

> i still dislike this for the broken progress reporting

If that eats up the benefit then we can abandon this. But then I would remove the composite Maven stage.

Hejado:
> No, it ran in the same executor.

In this case, LGTM.

> I still dislike this for the broken progress reporting

What exactly is broken here?

cdietrich:
give min 20 mins to restart build and take a screenshot

cdietrich:
see also https://github.com/eclipse/xtext/issues/1454
you see the parallel stage is green and shows 0 s

cdietrich:
normal progress reporting looks like

Bildschirmfoto vom 2019-07-02 12-15-19

cdietrich:
and here is the broken reporting

Bildschirmfoto vom 2019-07-02 12-28-19

cdietrich:
@Hejado did you get my point?

Hejado:
Yes, of course.

The logs show that both stages are working, so this seems to be a bug in the view alone...

Hejado:
Maybe we can try to create a MWE so we can file a bug against the 'full stage view' plugin (I think that's the name)

kthoms:
So shall we drop this for now?

kthoms:
Agreed with @cdietrich to drop this now

PR #905, opened 2019-08-13
[eclipse/xtext-lib#224] Stable method order when using Delegate 

szarnekow:
[eclipse/xtext-lib#224] Stable method order when using Delegate annotation


cdietrich:
did you check DelegateCompilerTest

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/DelegateCompilerTest.xtend
```diff
@@ -539,5 +539,44 @@ class DelegateCompilerTest extends AbstractXtendCompilerTest {
...
+				def void a()
+				def void b()
+			}
+			class C implements A {
+				@Delegate A a
+			}
+		'''
+		text.file.assertNoIssues
+		text.compile [
+			assertEquals('''
```
cdietrich: nice

Issue #751, opened 2019-03-15
Import static and static extension refactoring


Hanksha:
There is a feature in JDT `Source > Add Import (MOD1+SHIFT+M)` which adds as static import the static field or method that is selected. I use it quite often for constants or util methods etc. I think it could be nice to have in Xtend but in an improved way.

In addition to adding the static import for the selected element it will also refactor the whole file where necessary as well (and remove the unused import, which JDT doesn't do). The behavior would be extended for Xtend's extension methods.

This is a little preview of how I see it work:
![import_static](https://user-images.githubusercontent.com/6960133/54430702-55e44800-475f-11e9-8315-a5c7434d5ecf.gif)

Could be nice as a quick assist too.

If you guys think it would be a useful new feature I can continue working on it.

cdietrich:
https://github.com/eclipse/xtext/issues/2379

Hanksha:
Is there an existing API to rewrite a `XMemberFeatureCall` to a `XFeatureCall`?

From what I understand `Stuff.doSomething('test')` is a `XMemberFeatureCall` expression and after adding the static import and changing it to `doSomething('test')` it becomes a `XFeatureCall`. I could not find a way to do that by manipulating the model itself so I tried with textual modification instead.

What I did is:
- Find the static `XMemberFeatureCall` expression from the selection
- Compute the necessary import
- Compute the region of the `memberTargetCall` + `.` or `::` to remove (get the corresponding INode and find the next sibling for `.` or `::` using the `grammarElement`)
- Apply the changes using `DocumentRewriter` within a rewrite session

Am I on the right track?

szarnekow:
Yes, that sounds about right to me

Hanksha:
By the way I'm having a little issue with `org.eclipse.xtext.xbase.ui.document.DocumentRewriter`. I can't access its `RewritableImportSection importSection` to add an import. I can only access the import section if I call `documentRewriter.newSection(...).importSection` and since I'm updating all occurrences of the static method in the document it is duplicating the imports.

If I create my own `RewritableImportSection` and apply it to the document it complains about overlapping changes.

szarnekow:
When you are working on a patch, please feel free to adjust that according to your needs. The API is still provisional and we are open to recommendations.

Hanksha:
I'm almost done with this, just need to add more tests. I'll try to PR during the week.

cdietrich:
FIxed in 2.20

[...]

PR #914, opened 2019-09-04
[#751] Add Import Static and Import Static Extension commands

Hanksha:
Add _Import Static_ and _Import Static Extension_ commands #751 

The commands are accessible from the `Refactoring > Import Static | Import Static Extension` menu of the Xtend editor or via shortcut `M1+SHIFT+M` and `M1+SHIFT+ALT+M`.

The  _Import Static_ command will add as static import the selected method (if possible) and refactoring all occurrences in the document so that it uses the static method call.

Example:

importstatic

See `org.eclipse.xtend.ide.tests.refactoring.importer.StaticMethodImporterTest` for the behaviors.

The  _Import Static Extension_ command will add as static extension import the selected method (if possible) and refactoring all occurrences in the document so that it uses the static extension method call. It means it will move the first argument and put it as target of the method call.

Example:

importstaticextension

See `org.eclipse.xtend.ide.tests.refactoring.importer.StaticExtensionMethodImporterTest` for the behaviors. It has two ignored tests that fail due to an issue with the `ImportOrganizer` which removes the import because a static explicit call doesn't mark it as used.

Signed-off-by: Vivien Jovet <vivien.jovet@gmail.com>


Hanksha:
It seems I have some flaky tests, I tried to run the build locally via `mvn -f tycho-pom.xml clean install -Platest` but I'm getting this error `Unable to read repository at https://ci.eclipse.org/xtext/job/xtext-lib/job/master/lastStableBuild/artifact/build/p2-repository/content.xml. Received fatal alert: handshake_failure`

cdietrich:
Which java version is used in maven ?

Hanksha:
I'm using 3.6.0

cdietrich:
Java version not maven version

Hanksha:
Edited the comment, pressed enter my mistake.

cdietrich:
That version mit be too old for modern tls connections

cdietrich:
You should also try to run your test only by going into the test project and run

mvn clean install -Dtest=TestClassName

Hanksha:
So I should try with a more recent java 8 version or java 9 ?

cdietrich:
You should use the latest java 8. If you have concerns with Oracle Licensing you can Have Ooenjdk from the adoptopenjdk project

Hanksha:
Using adoptopenjdk8 fixed it, thanks.

cdietrich:
hmmm i did have a look at the code and have no clue why its failing. maybe we need to run this with some logging to see whats happening

cdietrich:
one thing that looks wired to me is the file names

```
def void testSingleStaticMethod() {
        'test.Test.xtend'.
```

guess that should be `test/Test.xtend`
am not sure if this is related to the problem.

cdietrich:
there are also some other invalid code places like

```
val i = 0
                    Other.doSo|mething(i)
                    Other.doSomething(i++)
```

Hanksha:
You're right, I'll double check the tests.

Hanksha:
I changed it to `test/Test.xtend` but a test that didn't fail before is now failing on jenkins (works locally).

cdietrich:
I cloud not reproduce this locally so far. Will try on Monday with xvnc

cdietrich:
the test / impl code does

``` val resource = document.priorityReadOnly[it]```
and then works with the resource. this looks bogus to me.
imho all the work on resource should happen within the "work"

**Comment by cdietrich** [comment]
@Hanksha i did some experiments and doing the 
`resourceCopier.loadIntoNewResourceSet(it)`
seems to help
https://github.com/eclipse/xtext-xtend/commit/2c127d714954bb93c39068205cd0645344e39a83
https://ci.eclipse.org/xtext/job/xtext-xtend/job/Hanksha-issue751/

**Comment by Hanksha** [comment]
@cdietrich Ok I'll try with that.

File: org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/XtendUiModule.xtend [hunk 1]
```diff
@@ -527,6 +529,14 @@ class XtendUiModule extends AbstractXtendUiModule {
 	def Class<? extends SyntaxColoringPreferencePage> bindSyntaxColoringPreferencePage() {
 		return XtendSyntaxColoringPreferencePage
 	}
+	
+	def Class<? extends StaticMethodImporter> bindStaticMethodImporter() {
+	    return StaticMethodImporter
```
cdietrich: why is this needed?

Hanksha: Is it not needed to be able to inject it ? Or does it automatically consider it a singleton?

cdietrich: its not really needed but it helps. singletons require a `@Singleton` annotation

Hanksha: So do you prefer to remove it here or annotate `StaticMethodImporter` and `StaticExtensionMethodImporter` classes with `@Singleton`?

cdietrich: you did me got wrong. that you did is not a singleton binding.

Hanksha: I understood that, my question is, should I remove it from here then?

cdietrich: no you can keep it

Hanksha:
@cdietrich Looks like the tests are not flaky anymore.

cdietrich:
yes the copy of the resources seems to have solved that issues. will try to find some time to go through the impl and tests later this week

Hanksha:
Alright, thanks for the help.

cdietrich:
Thx for the Contribution @Hanksha

PR #924, opened 2019-09-26
[https://github.com/eclipse/xtext/issues/1486]

miklossy:
- Use xtend32.png instead of modeling32.png for the feature image.


cdietrich:
looks like the feature branding in eclipse is broken there or isnt it?
i am not sure which thing counts with features.

plugin + about
or image in feature.xml.

lets test what is used.

miklossy:
I updated the build.properties to contain the `xtend32.png` file.

PR #944, opened 2019-10-30
[eclipse/xtext#1518] Fixed warning about missing entries in MANIFEST.MF

szarnekow:
[eclipse/xtext#1518] Fixed warning about missing entries in MANIFEST.MF


org.eclipse.xtend.ide.tests.data/.settings/org.eclipse.pde.prefs
```diff
@@ -0,0 +1,34 @@
+compilers.f.unresolved-features=1
```
szarnekow: This plugin has packages, that are not exported and Xtend tests rely on them being not exported. Thus I updated the compiler settings in this particular case

PR #950, opened 2019-11-06
[releng] Bootstrap against 2.20.0.M2

genie-xtext:
This change was brought to you by your friendly Xtext Genie.


kthoms:
Change in `org.eclipse.xtend.maven.plugin/src/test/resources/it/pom.xml` is missing, see https://github.com/eclipse/xtext-xtend/pull/910 for example

PR #962, opened 2019-11-19
fixed build.properties regarding example zips in contents generated by maven build

cdietrich:
fixed build.properties regarding example zips in contents generated by maven build


org.eclipse.xtend.examples/contents/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.zip
```
cdietrich: abuse as git keep. also the contents dir will be always here

Issue #919, opened 2019-09-10
Implement Quickfix 'Add JUnit library to the buildpath/classpath'


miklossy:
**JDT** offers a quickfix if in a Java Project the `org.junit.Test` is imported but the junit library is not found on the buildpath/classpath:
![screenshot](https://user-images.githubusercontent.com/20393472/64622821-35343c80-d3e8-11e9-978e-c67da174b73b.png)

The Xtend IDE should also offer a similar quickfix:
![screenshot2](https://user-images.githubusercontent.com/20393472/64622883-4bda9380-d3e8-11e9-94c7-d16311c7fbe3.png)

**Hint:** The Xtend IDE already offers a [quickfix](https://github.com/eclipse/xtext-xtend/blob/master/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/quickfix/XtendQuickfixProvider.java#L638) to add the Xbase lib to the classpath, a quickfix for adding the JUnit library could be implemented similarly.

Hanksha:
@miklossy What version of JDT are you using to get that quick fix? I tried with the 2019-09 release and I could not get it, maybe my setup is missing something

Hanksha:
JDT quick fix can add it both to the `.classpath` and the `META-INF/MANIFEST.MF`. 
![image](https://user-images.githubusercontent.com/6960133/67827616-7d7bfb00-fb0b-11e9-89e1-5462bf4dbe90.png)

For the latter should we use PDE to help with that (I'm guessing it's the PDE plugin that proposes that quick fix)? I noticed the `org.eclipse.xtend.ide` plugin doesn't have any dependency to PDE.

miklossy:
@Hanksha I would not use PDE here, but implement the Quickfix analog to the [XtendLibClasspathAdder](https://github.com/eclipse/xtext-xtend/blob/master/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/buildpath/XtendLibClasspathAdder.java). It already supports adding the missing library both to the classpath and to the MANIFEST.MF file.

Hanksha:
@miklossy Ok thanks I'll take a look at that.

Hanksha:
I've got a quick fix working for adding junit 4 or 5 to the classpath.
![add_junit4](https://user-images.githubusercontent.com/6960133/69214734-a9d2e680-0ba2-11ea-9a55-aa45516d8d69.gif)
[...]

PR #965, opened 2019-11-22
[#919] Add quickfixes that adds JUnit 4 or 5 libs to the classpath

Hanksha:
[#919] Add quickfixes that adds JUnit 4 or 5 libs to the classpath.

Similar to the quickfix that adds Xtend libs, it will first check if the
project is a plugin project and add the required bundles to the `MANIFEST.MF`, if not it
will add a container entry to the `.classpath`. The quickfixes appear for errors on
import declaration that starts by "org.junit" or
"org.junit.jupiter".

Examples:
junit4_quickfix
junit5_quickfix

What I haven't done yet is show the quickfix only if JUnit4/5 can "really" be added to the classpath. Currently it still proposes the quickfix even if after applying it the error will persist.

Signed-off-by: Vivien Jovet <vivien.jovet@gmail.com>

### DIFF
--- a/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/buildpath/Junit5LibClasspathAdder.java
+++ b/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/buildpath/Junit5LibClasspathAdder.java
@@ -1,0 +1,40 @@
+/*******************************************************************************
+ * Copyright (c) 2019 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtend.ide.buildpath;
+
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Path;
+import org.eclipse.jdt.core.IClasspathEntry;
+import org.eclipse.jdt.core.JavaCore;
+
+/**
+ * @author vivienjovet - Initial contribution and API
+ */
+public class Junit5LibClasspathAdder extends AbstractLibClasspathAdder {
+
+	public static final IPath JUNIT5_LIBRARY_PATH = new Path("org.eclipse.jdt.junit.JUNIT_CONTAINER/5");
+	public static final String[] BUNDLE_IDS = new String[] { 
+			"org.junit", "org.junit.jupiter.api", "org.junit.jupiter.engine",
+			"org.junit.jupiter.migrationsupport", "org.junit.jupiter.params", 
+			"org.junit.platform.commons", "org.junit.platform.engine",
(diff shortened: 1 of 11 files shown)


org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/buildpath/Junit5LibClasspathAdder.java
```diff
@@ -0,0 +1,40 @@
...
+import org.eclipse.jdt.core.IClasspathEntry;
+import org.eclipse.jdt.core.JavaCore;
+
+/**
+ * @author vivienjovet - Initial contribution and API
+ */
+public class Junit5LibClasspathAdder extends AbstractLibClasspathAdder {
+
+	public static final IPath JUNIT5_LIBRARY_PATH = new Path("org.eclipse.jdt.junit.JUNIT_CONTAINER/5");
+	public static final String[] BUNDLE_IDS = new String[] {
```
Hanksha: When I used the Java equivalent quickfix it added all those bundles.

org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/quickfix/XtendQuickfixProvider.java
```diff
@@ -174,9 +180,11 @@ public boolean hasResolutionFor(String issueCode) {
 			if(LINKING_ISSUE_CODES.contains(issue.getCode())){
 				List<IssueResolution> result = new ArrayList<IssueResolution>();
 				result.addAll(getResolutionsForLinkingIssue(issue));
-				return result;
-			} else
-				return super.getResolutions(issue);
+				if(!result.isEmpty())
```
Hanksha: Not sure if it's the right way to do it but without that it won't call the `fixJunitNotOnClasspath` method I added.

org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/buildpath/Junit5LibClasspathAdder.java
cdietrich: in the new Xtext project wizard we seem to have a different strategy.
we add org.junit.jupiter.api.Test only + the junit 5 container.

this might be cause tycho injects the engine from outside.
thus i wonder what we should do here.

cdietrich: the api dep seems to draw some of the others transitively anyway
Bildschirmfoto vom 2019-11-22 10-10-50

Hanksha:
@cdietrich We can create a separate issue for that. Updating the pom.xml would be easier than the build.gradle.

cdietrich:
yes as i have said the add xtend libary does the same (even if its a plugin project)

cdietrich:
thx @Hanksha

cdietrich:
@Hanksha could you Write a few sentences for the release notes

Hanksha:
@cdietrich Will do.

Issue #789, opened 2019-04-25
Xtend IDE: find *.xtend tests cases on the package level


miklossy:
Given is an Xtend Project with two test cases residing in the package `foo`:
```
package foo

import org.junit.Test
import static extension org.junit.Assert.assertEquals

class Foo1Test {
	
	@Test def test() {
		"bar".assertEquals("bar")
	}
}
```

```
package foo

import org.junit.Test
import static extension org.junit.Assert.assertEquals

class Foo2Test {
	
	@Test def test() {
		"foo".assertEquals("foo")
	}
	
}
```

The context menu `Run As -> JUnit Test` is missing wenn right clicking on the foo package:
![screenshot](https://user-images.githubusercontent.com/20393472/56714492-403a4900-6735-11e9-97c1-086d6eee3ef4.png)

miklossy:
See also https://bugs.eclipse.org/bugs/show_bug.cgi?id=449915

miklossy:
While working on this issue, I recognized a part in the [org.eclipse.xtend.ide/plugin.xml](https://github.com/eclipse/xtext-xtend/blob/master/org.eclipse.xtend.ide/plugin.xml#L1032-L1036) that looks bogus to me. The adatpter factory definition of IJavaElements the `adaptableType` and the `adapter type` is swapped. So I think the
```
     <factory 
        class="org.eclipse.xtend.ide.XtendExecutableExtensionFactory:org.eclipse.xtext.xbase.ui.launching.JavaElementDelegateAdapterFactory"
        adaptableType="org.eclipse.xtext.xbase.ui.launching.JavaElementDelegateJunitLaunch">
        <adapter type="org.eclipse.jdt.core.IJavaElement"/>
     </factory>
```
should be 
```
     <factory 
        class="org.eclipse.xtend.ide.XtendExecutableExtensionFactory:org.eclipse.xtext.xbase.ui.launching.JavaElementDelegateAdapterFactory"
        adaptableType="org.eclipse.jdt.core.IJavaElement">
        <adapter type="org.eclipse.xtext.xbase.ui.launching.JavaElementDelegateJunitLaunch"/>
     </factory>
```

kthoms:
I'm often confused with the direction of who needs an adapter of some type from whom. What makes you especially think that this is wrong?

cdietrich:
```
 * @param adaptableObject the adaptable object being queried
	 *   (usually an instance of <code>IAdaptable</code>)
[...]

PR #969, opened 2019-11-27
[#789] Find *.xtend test cases on the package level.

miklossy:
- Modify the plugin.xml of the Xtend IDE to ensure that the
JavaElementDelegateJunitLaunch is invoked when right clicking on a java
package.
- Modify the XtendJavaElementDelegateJunitLaunch to provide a proper
implementation of the getJavaElementForPackage(IPackageFragment) method
to deliver the corresponding java package in the xtend-gen folder.

### DIFF
--- a/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/launching/XtendJavaElementDelegateJunitLaunch.java
+++ b/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/launching/XtendJavaElementDelegateJunitLaunch.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2019, 2020 itemis AG (http://www.itemis.eu) and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -7,14 +7,25 @@
  *******************************************************************************/
 package org.eclipse.xtend.ide.launching;
 
+import java.util.Set;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Path;
 import org.eclipse.emf.common.util.EList;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.jdt.core.IJavaElement;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.IPackageFragment;
+import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.xtend.core.xtend.XtendFile;
 import org.eclipse.xtend.core.xtend.XtendTypeDeclaration;
+import org.eclipse.xtext.builder.EclipseOutputConfigurationProvider;
 import org.eclipse.xtext.common.types.JvmIdentifiableElement;
 import org.eclipse.xtext.common.types.util.jdt.IJavaElementFinder;
+import org.eclipse.xtext.generator.OutputConfiguration;
 import org.eclipse.xtext.parser.IParseResult;
+import org.eclipse.xtext.resource.FileExtensionProvider;
 import org.eclipse.xtext.resource.XtextResource;
 import org.eclipse.xtext.xbase.ui.launching.JavaElementDelegateJunitLaunch;
 
@@ -28,6 +39,12 @@ public class XtendJavaElementDelegateJunitLaunch extends JavaElementDelegateJuni
 	@Inject
 	private IJavaElementFinder elementFinder;
 
+	@Inject
+	private EclipseOutputConfigurationProvider outputConfigurationProvider;
+
+	@Inject
+	private FileExtensionProvider fileExtensionProvider;
+
 	@Override
 	protected IJavaElement findJavaElement(XtextResource resource, int offset) {
 		IJavaElement javaElement = super.findJavaElement(resource, offset);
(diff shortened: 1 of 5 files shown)


cdietrich (changes requested):
this seems to destry the run on the file and within the file (rightclick)

miklossy:
You mean that the JUnit Menu is not there at all?

cdietrich:
Yes no entries at all

miklossy:
@cdietrich you were right, the menus indeed have been disappeared. I updated the PR, now it should be fine. Please check again. Thank you!

cdietrich:
@miklossy selecting a source folder is still not working.

miklossy:
@cdietrich Currently, when selecting a source folder, the JUnit Menu (provided by JDT) appears, but selecting it reports that no JUnit test cases are found.

Making the `org.eclipse.jdt.core.IPackageFragmentRoot` (similar to the `org.eclipse.jdt.core.IPackageFragment` class) adaptable, would results that two JUnit menus will be present (one will work, but the one provided by JDT will not).

cdietrich:
@miklossy  why dont i get two entries for packages then?

miklossy:
@cdietrich JDT offers no JUnit Menu on packages containing no java files. The Xtend IDE offers the JUnit Menu only on packages containing no java files but at least one `*.xtend` file. See `boolean containsOnlyXtendFiles(IPackageFragment packageFragment)` method in the PR.

cdietrich:
No I mean a package with one xtend and one java file

miklossy:
When this PR is applied, the JUnit Menu is offered on packages containing no java files but at least one *.xtend file, see also [screenshot](https://user-images.githubusercontent.com/20393472/56714492-403a4900-6735-11e9-97c1-086d6eee3ef4.png) attached to the issue eclipse/xtext#2283

miklossy:
If you remove the containsOnlyXtendFiles(packageFragment) guard from the PR than you will get two JUnit menu on packages containing both java and xtend files:
screenshot

cdietrich:
yes but what i dont understand: why cant the very same concept not be ported to src folders?
cause jdt alsways shows something. even if there is no java file?

miklossy:
JDT shows nothing on packages containing no java files. But JDT shows the JUnit menu point on source folders even if they contain no java files.

cdietrich:
:(

cdietrich:
i really hate what jdt does here

cdietrich:
@szarnekow what do you think?

miklossy:
We should investigate the use case with the source folders later on. Currently for me the use case on packages is the important one.

cdietrich:
behaviour is different between junit 4 (executes java only) and 5 (executes both)
and between packages and src folders

szarnekow:
Is there already a consensus on what we want to achieve?
I'd vote for running the Xtend tests and the Java tests when selecting either a source folder or a package from a source folder.
If a xtend-gen or package from xtend-gen is selected, we should probably not try to outsmart anything and just do what JDT does in those cases.
Junit4 and Junit5 should behave identically.

miklossy:
> 
> 
> Is there already a consensus on what we want to achieve?
The goal of this PR is to improve the Xtend IDE behavior on packages containing only `*.xtend` files.
> I'd vote for running the Xtend tests and the Java tests when selecting either a source folder or a package from a source folder.
JDT behaves differently when selecting a source folder and a package containing no java test cases:
- On source folder selection, JDT already provides a JUnit menu, when clicking that, there is a message that no JUnit test cases are found. If the Xtend IDE would provide its own JUnit menu, there would be two JUnit menus offered.
- On package selection, JDT does not provide any menu, so that the Xtend IDE can provide its own JUnit menu.
> If a xtend-gen or package from xtend-gen is selected, we should probably not try to outsmart anything and just do what JDT does in those cases.
- Its already happens, see the `containsOnlyXtendFiles` guard.
> Junit4 and Junit5 should behave identically.
There is a different behaviour on packages containing both `*.java` and `*.xtend` files.
- In case of Junit4, only the test cases implemented in Xtend will be executed.
- In case of Junit5, both test cases will be executed.
I like the Junit5 behaviour here, but to make that also work for Junit4, the JDT behaviour should be improved here, I guess.

cdietrich:
@miklossy i cannot follow you: here is was JDT does on package

junit_on_package

miklossy:
@cdietrich I mean if you selected a package containing **no java test cases**, there is no JDT JUnit menu offered.

cdietrich:
yes but i wonder how this is calculated and if we can take part there.
my expectation is:

if there is a  java junit test and a xtend unit test both get executed

miklossy:
> 
> 
> yes but i wonder how this is calculated and if we can take part there.
> my expectation is:
> 
> if there is a java junit test and a xtend unit test both get executed

This should already be the case for `Junit5`, right? only `Junit4` behaves different. Can you confirm that?

cdietrich:
will do when i find the time to examine

cdietrich:
@szarnekow @miklossy how to proceed?

miklossy:
I would like to merge this as soon as possible and collect feedback from the community afterwards.

org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/launching/XtendJavaElementDelegateJunitLaunch.java
```diff
@@ -38,6 +44,65 @@ protected IJavaElement findJavaElement(XtextResource resource, int offset) {
...
+			}
+		}
+
+		return false;
+	}
+
+	private IJavaElement getXtendGenPackageFragment(IPackageFragment packageFragment) {
+		String packageName = packageFragment.getElementName();
+		IJavaElement project = packageFragment.getAncestor(IJavaElement.JAVA_PROJECT);
+		if (project instanceof IJavaProject) {
```
szarnekow: This should use the outputconfig instead of hard coded paths

szarnekow (changes requested):

org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/launching/XtendJavaElementDelegateJunitLaunch.java
miklossy: Thanks for the feedback @szarnekow ! I udpated the PR to use the `org.eclipse.xtext.builder.EclipseOutputConfigurationProvider` instead of the hard-coded path.

org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/launching/XtendJavaElementDelegateJunitLaunch.java
```diff
@@ -38,6 +51,77 @@ protected IJavaElement findJavaElement(XtextResource resource, int offset) {
...
+					return null;
+				}
+				return xtendGenPackageFragment;
+			}
+		}
+		return null;
+	}
+	
+	private String getXtendOutputFolder(String packageName) {
+		Set<OutputConfiguration> outputConfigurations = outputConfigurationProvider.getOutputConfigurations();
```
szarnekow: Instead of using the first output folder, I'd recommend to return the output folder, that matches the original package fragment root.

miklossy:
I extended the PR with corresponding `XtendJavaElementDelegateJunitLaunchTest` test
cases.

org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/launching/XtendJavaElementDelegateJunitLaunch.java
miklossy: As far as I can see, the `OutputConfiguration.getOutputDirectory` method returns only the name of the output folder (in this case `xtend-gen`), so there is nothing with that I could compare the original package fragment root. I would suggest to keep that implementation as it is and improve that later on on demand.

szarnekow: As far as I can tell, this will break immediately if a Xtend is used in a maven project with `src/main/java` and `src/test/java` source folders. You may want to take inspiration from the org.eclipse.xtend.ide.builder.SourceRelativeFileSystemAccess

miklossy:
@szarnekow You are absolutly right, unforunatly this solution does not work for maven project with `src/test/java` source folders:
screenshot

PR #978, opened 2019-12-12
[977] Further refinement of the checkJUnitMethodReturnType Xtend validation…

sailingKieler:
… rule incorporating the test runner setting.

Signed-off-by: Christian Schneider <christian.schneider@typefox.io>

### DIFF
--- a/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/JUnitMethodReturnTypeValidationTest.xtend
+++ b/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/JUnitMethodReturnTypeValidationTest.xtend
@@ -11,8 +11,8 @@ import javax.inject.Inject
 import org.eclipse.emf.ecore.EClass
 import org.eclipse.xtend.core.tests.AbstractXtendTestCase
 import org.eclipse.xtend.core.xtend.XtendFile
-import org.eclipse.xtext.testing.util.ParseHelper
 import org.eclipse.xtext.testing.validation.ValidationTestHelper
+import org.junit.FixMethodOrder
 import org.junit.Test
 
 import static org.eclipse.xtend.core.validation.IssueCodes.INVALID_RETURN_TYPE_IN_CASE_OF_JUNIT_ANNOTATION
@@ -20,10 +20,10 @@ import static org.eclipse.xtend.core.xtend.XtendPackage.Literals.*
 import static org.eclipse.xtext.diagnostics.Diagnostic.LINKING_DIAGNOSTIC
 import static org.eclipse.xtext.xbase.XbasePackage.Literals.*
 
+@FixMethodOrder(JVM)
 class JUnitMethodReturnTypeValidationTest extends AbstractXtendTestCase {
 
 	@Inject extension ValidationTestHelper
-	@Inject extension ParseHelper<XtendFile>
 
 	@Test def test001() {
 		'''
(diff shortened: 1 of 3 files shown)


sailingKieler:
See also eclipse/xtext#2335.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/validation/XtendValidator.java
```diff
@@ -361,6 +368,32 @@ private boolean hasJUnitAnnotation(JvmOperation operation) {
 		return !annotationUtil.findAnnotations(junitAnnotations, operation).isEmpty();
 	}
 	
+	protected boolean classUsesCustomRunner(XtendTypeDeclaration clazz) {
```
szarnekow: Since `@RunWith` is also `@Inherited`, we need to check the super type hierarchy.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/validation/XtendValidator.java
```diff
@@ -361,6 +368,32 @@ private boolean hasJUnitAnnotation(JvmOperation operation) {
...
+		final XtextResource res = (XtextResource) clazz.eResource();
+		return res.getCache().get("usesCustomRunner" + clazz.toString(), res, () -> {
+			final EObject jvmElement = jvmModelAssociations.getPrimaryJvmElement(clazz);
+			if (jvmElement instanceof JvmDeclaredType) {
+				final Iterable<JvmAnnotationReference> annRefs = filter(
```
szarnekow: I think imperative coding style would greatly help to follow the logic here. But that's a matter of taste.

dhuebner (approved):

org.eclipse.xtend.core/src/org/eclipse/xtend/core/validation/XtendValidator.java
szarnekow: This looks like `XAnnotationUtil.findAnnotation` or `org.eclipse.xtext.common.types.util.AnnotationLookup.findAnnotation(JvmAnnotationTarget, String)`

org.eclipse.xtend.core/src/org/eclipse/xtend/core/validation/XtendValidator.java
```diff
@@ -339,7 +346,7 @@ public void checkJUnitMethodReturnType(XtendFunction function) {
 		 * Active annotations could also change the return type.
 		 * Checking that the JvmOperation really has a JUnit annotation.
 		 */
-		if(hasJUnitAnnotation(operation)) {
+		if(!classUsesCustomRunner(function.getDeclaringType()) && hasJUnitAnnotation(operation)) {
```
szarnekow: This should probably check operation.getDeclaringType rather than function.getDeclaringType

szarnekow (changes requested):

org.eclipse.xtend.core/src/org/eclipse/xtend/core/validation/XtendValidator.java
sailingKieler: right, makes sense.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/validation/XtendValidator.java
sailingKieler: true!

org.eclipse.xtend.core/src/org/eclipse/xtend/core/validation/XtendValidator.java
sailingKieler: Good point, thanks!

sailingKieler: @szarnekow is there any helper method providing all the super types?

szarnekow: Yes, there is `org.eclipse.xtext.common.types.util.RawSuperTypes`

sailingKieler: right, Thanks. Didn't work with Xbase/the types stuff for quite some time...

sailingKieler:
Pushed update corresponding to your remarks.

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/JUnitMethodReturnTypeValidationTest.xtend
```diff
@@ -444,22 +444,552 @@ class JUnitMethodReturnTypeValidationTest extends AbstractXtendTestCase {
...
+			@RunWith(com.foo.bar.MyRunner)
+			class Foo {
+				
+				@Test def test() {
+				}
+			}
+		'''.withRunner('''
+			package com.foo.bar
+
+			class MyRunner extends org.eclipse.xtext.testing.XtextRunner {
```
szarnekow: I'm not sure about this test expectation. `MyRunner` extends (transitively) `org.junit.runners.BlockJUnit4ClassRunner` which has `org.junit.runners.BlockJUnit4ClassRunner.validateTestMethods(List<Throwable>)` and will throw if a test method is not `void`.

szarnekow (changes requested):
I don't think we'll find an easy to understand consensus on the validation. E.g. this change will miss all subtypes of BlockJunit4ClassRunner.
Maybe we should rather make the validation itself configurable such that users who implemented own runners can disable it?
Or it should become smart enough and closer to the runtime semantics of Junit4.

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/JUnitMethodReturnTypeValidationTest.xtend
sailingKieler: Correct. But If I add an override like
```
override protected validatePublicVoidNoArgMethods(Class<? extends Annotation> annotation, boolean isStatic, List<Throwable> errors) {
}
```
it doesn't.

To be honest, I don't see why Xtend shall enforce the defaults of a library. Is there any other library out there for which you would accept that?
I can see the point that a compile time check is preferable over a runtime check in practice, and that's why the checks are still active in the default case.

sailingKieler: (and for completeness:) ...or
```
override protected collectInitializationErrors(List<Throwable> errors) {
}
```

szarnekow: I understand what you are saying. But please also consider the first sentence of the `org.junit.Test` annotation's Javadoch: `The Test annotation tells JUnit that the public void method to which it is attached can be run as a test case.` So I'm not sure that whatever test runner you are using is violating the contract of the Test annotation.

miklossy: Making the validation severity configurable could be the right way to go as suggested by @dhuebner in his previous comment: https://github.com/eclipse/xtext-xtend/issues/516#issuecomment-518571016

sailingKieler: > I understand what you are saying. But please also consider the first sentence of the `org.junit.Test` annotation's Javadoch: `The Test annotation tells JUnit that the public void method to which it is attached can be run as a test case.` So I'm not sure that whatever test runner you are using is violating the contract of the Test annotation.

That's indeed true ... hmmm.

cdietrich:
i have no clue how to procede with this

szarnekow:
I think making the validation configurable would be a good compromise. In the current shape, I'd not approve this PR.

miklossy:
> 
> 
> I think making the validation configurable would be a good compromise. In the current shape, I'd not approve this PR.

@sailingKieler Could you please update the PR as @szarnekow and @dhuebner suggested?

miklossy (changes requested):
I would also prefer to make the validation rule configurable, so that everyone can configure that according to his/her custom needs.

sailingKieler:
Although that I'm still convinced that it shouldn't be Xtend's job to implement and enforce semantics of any framework's client API, the remark in https://github.com/eclipse/xtext-xtend/pull/978#discussion_r358172450 convinced me to better switch to a custom marker annotation.
Consequently, I close this PR.

PR #979, opened 2019-12-13
[eclipse/xtext-eclipse#1228] Impl getDocComment to use it in wizard.

ArneDeutsch:

### DIFF
--- a/org.eclipse.xtend.core/src/org/eclipse/xtend/core/macro/declaration/CompilationUnitImpl.xtend
+++ b/org.eclipse.xtend.core/src/org/eclipse/xtend/core/macro/declaration/CompilationUnitImpl.xtend
@@ -155,7 +155,11 @@ class CompilationUnitImpl implements CompilationUnit {
 	}
 
 	override getDocComment() {
-		throw new UnsupportedOperationException("Auto-generated function stub")
+		val nodes = fileHeaderProvider.getFileHeaderNodes(xtendFile.eResource)
+		if (nodes.empty)
+			return ""
+		else
+			return nodes.get(0).text
 	}
 
 	override getPackageName() {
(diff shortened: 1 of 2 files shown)


org.eclipse.xtend.core/src/org/eclipse/xtend/core/macro/declaration/CompilationUnitImpl.xtend
```diff
@@ -155,7 +155,11 @@ class CompilationUnitImpl implements CompilationUnit {
 	}
 
 	override getDocComment() {
-		throw new UnsupportedOperationException("Auto-generated function stub")
+		val nodes = fileHeaderProvider.getFileHeaderNodes(xtendFile.eResource)
+		if (nodes.empty)
+			return ""
```
szarnekow: Should this fallback to `getFileHeader`?

ArneDeutsch: No. `getFileHeader` returns only the content of the comment. Inserting that would break the java code.

szarnekow: Yes, I'm aware of that. The question was more in the direction: should we fallback to `"/*" + getFileHeader + "*/"`

ArneDeutsch: To my understanding `getFileHeader` will return nothing usable (even `null` is possible) in case there are no nodes. Because of this I do not see an advantage but even more possibilities for errors.

szarnekow: Ok

PR #980, opened 2019-12-16
Refactor usage of Wrapper constructors to valueOf

kthoms:
The constructors of Wrapper classes have been deprecated. It is
recommended to use their static valueOf() method. For Boolean the
literals TRUE and FALSE can be used.

### DIFF
--- a/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/AbstractXbaseEvaluationTest.java
+++ b/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/AbstractXbaseEvaluationTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2016 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2010, 2019 itemis AG (http://www.itemis.eu) and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -643,7 +643,7 @@ public abstract class AbstractXbaseEvaluationTest extends Assert {
 	}
 
 	@Test public void testUnaryOperator_00() throws Exception {
-		assertEvaluatesTo(new Integer(-19),"-19");
+		assertEvaluatesTo(Integer.valueOf(-19),"-19");
 	}
 	
 	@Test public void testUnaryOperator_01() throws Exception {
@@ -663,7 +663,7 @@ public abstract class AbstractXbaseEvaluationTest extends Assert {
 	}
 	
 	@Test public void testUpToOperator() throws Exception {
-		assertEvaluatesTo(new Integer(9),"(9..13).iterator().next()");
+		assertEvaluatesTo(Integer.valueOf(9),"(9..13).iterator().next()");
 	}
 	
 	@Test public void testExclusiveRangeOperator_0() throws Exception {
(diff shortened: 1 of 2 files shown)


org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/AbstractXbaseEvaluationTest.java
```diff
@@ -3640,44 +3640,44 @@
 	}
 	
 	@Test public void testMemberCallOnMultiType_02() throws Exception {
-		assertEvaluatesTo(Integer.valueOf(20), "(if (false) new Double('-20') else new Integer('20')).intValue");
-		assertEvaluatesTo(Integer.valueOf(-20), "(if (true) new Double('-20') else new Integer('20')).intValue");
+		assertEvaluatesTo(Integer.valueOf(20), "(if (false) new Double('-20') else Integer.valueOf('20')).intValue");
```
szarnekow: Not sure if this change here was intentional. Either we also use Double.valueOf(..) or we stick to the constructor for Integer, too

szarnekow: There appears to be a pattern of similar things. Isn't Double(String) deprecated, too?

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/AbstractXbaseEvaluationTest.java
```diff
@@ -2662,7 +2662,7 @@
 	}
 	
 	@Test public void testConstructor_02() throws Exception {
-		assertEvaluatesTo(new Character('a'), "new Character('abc'.charAt(0))");
+		assertEvaluatesTo(Character.valueOf('a'), "Character.valueOf('abc'.charAt(0))");
```
szarnekow: The test is named 'testConstructor' - after this change there is no constructor involved anymore.

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/AbstractXbaseEvaluationTest.java
kthoms: True. I have only refactored the occurrences where the constructors where used in a static context. There are other occurrences, e.g. in test expectations in rich strings.
Shall they all be refactored?

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/AbstractXbaseEvaluationTest.java
kthoms: I was too eager here. May we use a different constructor here for testing? Or revert this one?

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/AbstractXbaseEvaluationTest.java
szarnekow: I didn't get the remark about the static context. In this line, we do have `new Double` and `new Integer` in the xbase code, and only one of both was updated. I have a slight preference towards keeping the xbase tests (that is, the string literals with the xbase expression) as is and only refactor the expression in the Java code.

kthoms: I missed that I have used valueOf in the xbase expression. This was not intended here.

PR #987, opened 2020-01-07
[eclipse/xtext#1455] Fix line endings for JavaConverterTest.

ArneDeutsch:


ArneDeutsch:
That's quite strange. Must be some subtle difference in line ending handling on our machines. If I do as you suggest the test fails for me. What if we call 'toUnixLineSeperator' in both places. Works then for me.

miklossy:
Adding the `toUnixLineSeparator` to both places results in a failing test on my machine.
image

cdietrich:
why is there windows endings in the expectation if you convert them?

miklossy:
That is the point. Currently we convert only the line endings of the actual parameter (in 2 places), but we does not convert the line endings of the expected parameter. IMHO we should convert both of them to keep consistency.

PR #1024, opened 2020-03-13
made tests java 14 ready

cdietrich:
made tests java 14 ready


org.eclipse.xtend.core/src/org/eclipse/xtend/core/javaconverter/ASTParserFactory.xtend
```diff
@@ -51,6 +51,8 @@ class ASTParserFactory {
 			case "12" : 11
 			// TODO update to 13 later
 			case "13" : 11
+			// TODO update to 14 later
```
szarnekow: Not sure about the 11's here in the long run.

cdietrich: sure. but we did the same for 12 and 13.

PR #1051, opened 2020-05-15
[#1050] Xtend precompile step should not blindly run annotations processors by default

### LINKED ISSUE #1050: Maven Xtend runs Java annotation processor in wrong source directory
We have a (Eclipse plugin) project with both an annotation processor and Xtend code.
This runs fine in Eclipse, but when we run the build with Maven (Tycho), an unexpected folder appears in the root build folder with a copy of all the generated sources by the annotation processor.

My hypothesis is that Xtend runs the annotation with the wrong "generated sources directory". But maybe the annotation processor does not have to run at all by Xtend?

szarnekow:
Allow to pass arguments to the pre built Java compiler. By default disable annotation processing.

closes #1050

### DIFF
--- a/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
+++ b/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
@@ -167,6 +167,12 @@ public class XtendBatchCompiler {
 	protected boolean writeStorageFiles = false;
 	private GeneratorConfig generatorConfig = new GeneratorConfig();
 	protected ClassLoader currentClassLoader = getClass().getClassLoader();
+	/**
+	 * Additional arguments to the Eclipse batch compiler for the precompile step.
+	 * Defaults to '-proc:none' to disable annotation processing. Clients may override.
+	 * @since 2.22
+	 */
+	protected String additionalPreCompileArgs = "-proc:none";
 
 	private URI baseURI;
 
@@ -214,6 +220,20 @@ public class XtendBatchCompiler {
 	 */
 	public void setWriteStorageFiles(boolean writeStorageFiles) {
 		this.writeStorageFiles = writeStorageFiles;
+	}
+	
+	/**
+	 * @since 2.22
+	 */
+	public void setAdditionalPreCompileArgs(String additionalPreCompileArgs) {
+		this.additionalPreCompileArgs = additionalPreCompileArgs;
+	}
+	
+	/**
+	 * @since 2.22
+	 */
+	public String getAdditionalPreCompileArgs() {
+		return additionalPreCompileArgs;
 	}
 	
 	@Inject
@@ -654,6 +674,9 @@ public class XtendBatchCompiler {
 		commandLine.add("-proceedOnError");
 		if (encodingProvider.getDefaultEncoding() != null) {
 			commandLine.add("-encoding \"" + encodingProvider.getDefaultEncoding() + "\"");
+		}
+		if (additionalPreCompileArgs != null) {
+			commandLine.add(additionalPreCompileArgs);
 		}
 		List<String> sourceDirectories = newArrayList(sourcePathDirectories);
 		commandLine.add(concat(" ", transform(sourceDirectories, new Function<String, String>() {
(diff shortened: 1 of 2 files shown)


org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
```diff
@@ -167,6 +167,12 @@ public boolean accept(File pathname) {
 	protected boolean writeStorageFiles = false;
 	private GeneratorConfig generatorConfig = new GeneratorConfig();
 	protected ClassLoader currentClassLoader = getClass().getClassLoader();
+	/**
+	 * Additional arguments to the Eclipse batch compiler for the precompile step.
+	 * Defaults to '-proc:none' to disable annotation processing. Clients may override.
+	 * @since 2.22
+	 */
+	protected String additionalPreCompileArgs = "-proc:none";
```
cdietrich, line 175: the assumtion is we dont break anyone?

szarnekow: Indeed.

People who had annotation processors on the classpath of their maven project would suffer from code being created directly in `${basedir}` rather than `target/something/something/gen`

szarnekow: Or do you mean we should apply the default only from mvn?

cdietrich: so we "break" already broken code only, so having it disabled by default wont be an issue?

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
```diff
@@ -655,6 +675,9 @@ protected boolean preCompile(File classDirectory, Iterable<String> sourcePathDir
 		if (encodingProvider.getDefaultEncoding() != null) {
 			commandLine.add("-encoding \"" + encodingProvider.getDefaultEncoding() + "\"");
 		}
+		if (additionalPreCompileArgs != null) {
```
kthoms, line 678: They can't be null, do they? Maybe instead of checking null here, do it in the setter?

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
kthoms, line 175: You have accessors, so maybe private?

kthoms: For precompiling I think disabling is better than using Maven defaults

szarnekow: I prefer consistency with the other fields in this class.

szarnekow: Well of course this argument is ridiculous in this class given the mixture of private / protected / accessors.
Any preferences?

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
szarnekow: I'd like people to be able to set null if they somehow *want* that.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
kthoms: I prefer consistency with the existing code style, so fine to me.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
kthoms: I can't imagine a meaningful use case for that. But since the field is protected and setter unguarded, it is safe to do the guard here. Let it be.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
szarnekow: Indeed, that would be my assumption.

cdietrich:
what is the state of this PR?

szarnekow:
Waiting for your approval.

Issue #1060, opened 2020-05-28
Add new Xtend editor quickfixes on *.xtend files


miklossy:
The Xtend IDE should provide the following quickfixes on `*.xtend` files:

- [#398023](https://bugs.eclipse.org/bugs/show_bug.cgi?id=398023) 
- [#412593](https://bugs.eclipse.org/bugs/show_bug.cgi?id=412593)
- [#454088](https://bugs.eclipse.org/bugs/show_bug.cgi?id=454088)

@loradd Would you like to take care of this issue?

loradd:
Yes, I can take care of this. Thanks!

loradd:
@miklossy it is not clear what is meant with constant in [#454088](https://bugs.eclipse.org/bugs/show_bug.cgi?id=454088), do we mean a final static instance member?

miklossy:
Yes

miklossy:
Currently, whenever the Xtend IDE recognizes that a method is missing, it offers a quickfix for creating the method:
![screenshot](https://user-images.githubusercontent.com/20393472/83997838-9aa95f80-a95f-11ea-8ff0-94728c2b3719.png)
It should be useful to offer an additional quickfix for creating a private method, in the example above there should be three quickfixes:

- Change to 'a'
- Create public method 'm'
- Create private method 'm'

@loradd Maybe you can fix this first.

szarnekow:
@miklossy This should be discussed. I don't want to have quickfixes for all permutations of modifiers.

miklossy:
@szarnekow I do not want to have quickfixes for all permutations of modifiers either. But creating a private method would be very useful, at least I would use that one more often than creating a public method. Currently I use the quickfix that creates a public method and have to modify that one to convert the created public method to private.

Let's see what others think about this.

kthoms:
Creating a private method sounds as a likely choice. If this would be added I would not change the text for creating a public method, though.

miklossy:
**Hint:** JDT offers a quickfix for creating a private method, but no quickfix to create a public one.
![screenshot2](https://user-images.githubusercontent.com/20393472/84002932-84a09c80-a969-11ea-8949-398242722e48.png)

szarnekow:
JDT uses placeholders in the proposed method.

miklossy:
@szarnekow It is true:
[...]

PR #1068, opened 2020-06-14
Add new Xtend editor quickfixes on *.xtend files #1060 

loradd:


org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/QuickfixTest.xtend
```diff
@@ -427,7 +427,23 @@ class QuickfixTest extends AbstractXtendUITestCase {
 			}
 		''')
 		.assertFeatureCallLinkingIssue
-		.assertResolutionLabels("Create field 'bar'", "Create local variable 'bar'", "Create method 'bar()'", "Create method 'getBar()'")
+		.assertResolutionLabels(
+			"Create constant 'bar'",
```
szarnekow, line 431: My association with `constant` would be a static final field.
If we create a final field but don't add a constructor, the model would be in worse shape than before the quick fix, wouldn't it?
I'm not a big fan.

org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/quickfix/CreateMemberQuickfixes.java
```diff
@@ -251,15 +260,21 @@ protected String getOperatorMethodName(XAbstractFeatureCall call) {
...
 	}
 	
-	protected void newLocalVariableQuickfix(final String variableName, XAbstractFeatureCall call, Issue issue,
+	protected void newLocalVariableQuickfixes(final String variableName, XAbstractFeatureCall call, Issue issue,
+			IssueResolutionAcceptor issueResolutionAcceptor) {
+		newLocalVariableQuickfix(variableName, false, call, issue, issueResolutionAcceptor);
+		newLocalVariableQuickfix(variableName, true, call, issue, issueResolutionAcceptor);
+	}
+	
+	protected void newLocalVariableQuickfix(final String variableName, boolean isImmutable, XAbstractFeatureCall call, Issue issue,
```
szarnekow, line 269: please rename `isImmutable` to `isFinal` since immutability offers other guarantees than a mere final variable.

loradd: Ok, done!

org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/QuickfixTest.xtend
loradd: I have added a default value literal initialisation for constants in _FieldBuilders.xtend_.

cdietrich:
thx @loradd

Issue #903, opened 2019-08-10
@Accessors does not generate @Override or @Deprecated when required


ivaniesta14:
Consider the following code:
```xtend
interface IFoo {
  abstract def Bar getBar()
}
class Baz implements IFoo {
  @Accessors Bar bar
}
```
In the compiled java class Baz, the method `getBar` is not tagged with `@Override`, triggering a warning.
I suspect this may also happen to the `@Data` annotation.

cdietrich:
the processors currently dont check and add override annotation

ivaniesta14:
Another related problem with the same cause: 
Marking the field as `@Deprecated` does not make the accessors deprecated, only the field.

cdietrich:
@ivaniesta14 do you consider a special treatment for `@Deprecated` or do you have any criteria which annotations should be pulled over and which ones not

ivaniesta14:
Ideally, the Java standard annotations (`@Deprecated`, `@Override` and `@SuppressWarnings`) and some nullness-related ones (notably `edu.umd.cs.findbugs.annotations.NonNull` and `Nullable` from Find/Spotbugs, `javax.annotation.Nonnull` and `CheckForNull` from JSR-305, and `org.eclipse.jdt.annotation.NonNull` and `Nullable` from Eclipse annotations) should be passed to the methods.
Alternatively, copying all annotations valid for both fields and methods is another option, but I think it may be inconvenient in some edge cases.

szarnekow:
@ivaniesta14 I added an API that allows to ask for the overridden methods. Do you want to give it a try and file PR with the API incorporated into the active annotation compiler?

ivaniesta14:
I would like to but I can't properly set up the project in Eclipse. I can commit the uncompiled (updated) Xtend code if needed.

ivaniesta14:
Based on the `@Accessors` annotation and its processor I managed to create a temporary workaround:
```xtend
@Active(AccessorsOverrideProcessor)
@Target(ElementType.FIELD)
@Documented
annotation AccessorsOverride {
	AccessorType[] value = #[AccessorType.PUBLIC_GETTER, AccessorType.PUBLIC_SETTER]
	/**Generate {@code @Override} annotation on getter */
	boolean overrideGetter = false
	/**Generate {@code @Override} annotation on setter */
	boolean overrideSetter = false
	/**Annotations to delegate to the getter */
	Class<? extends Annotation>[] delegateGetter = #[]
[...]

PR #1074, opened 2020-06-22
[#903] Allow to ask methods for their base method

szarnekow:
[#903] Allow to ask methods for their base method


org.eclipse.xtend.core/src/org/eclipse/xtend/core/macro/declaration/XtendBasedDeclarations.xtend
```diff
@@ -339,6 +334,10 @@ class XtendMethodDeclarationImpl extends XtendMemberDeclarationImpl<XtendFunctio
 		delegate.parameters.map[compilationUnit.toXtendParameterDeclaration(it)]
 	}
 	
+	override getOverriddenOrImplementedMethods() {
+		emptyList
```
cdietrich, line 338: no override for XtendMembers?

szarnekow: It would be the only place that I can spot where we go from Xtend->Associations->Jvm.
I'll do a follow-up PR.

Issue #1061, opened 2020-05-30
Provide ui test infrastructure for testing multi quickfixes [2/3]


drkstr101:
This is a follow up on: https://github.com/eclipse/xtext-eclipse/issues/1447

As the next step will take place in this repository, I am opening up an issue to create the `EqualsWithNullMultiQuickfixTest` and `RemoveUnnecessaryModifiersMultiQuickfixTest` test cases described in the referenced issue above.

miklossy:
@drkstr101 Are you still working on this issue? Please let us know if you cannot manage, in that case I can take care of this.

drkstr101:
@miklossy I tried working on this last week but I kept getting install errors when attempting to setup 2020-06 contributors bundle. I was thinking maybe it was just overloaded due to the recent release, so I will try again now and see if the weekend makes any difference. I will post back once I'm up and running again. I have should have this wrapped up soon after that.

cdietrich:
its still broken right now because https://github.com/eclipse/xtext-eclipse/pull/1479 does not get reviewed

drkstr101:
Will it cause any problems if I just continue to build off 2020-03?

I still get:

    No repository found at https://download.eclipse.org/eclipse/updates/4.16-I-builds.

On 2020-06 Install.

cdietrich:
Yes. If you switch to my branch for Xtext-eclipse it should work

drkstr101:
Yeah I still got the same error with 2020-03, so let me give that I try.

Thanks for the help!

drkstr101:
@miklossy My apologies for being AFK about a week. I was able to take a bit of personal time though, and am all refreshed and ready to tackle this again. I am now up and running on `2020-06` without issue, and have a totally free schedule this week. My current mission is to finish out this task or bust!

drkstr101:
When I attempt to run `xtend.ide.tests.slow (xtend).launch`, I am getting:

    The input element of the launch configuration does not exist

I was unable to locate any additional context. The other launchers work as expected.

My understanding is that my tests should go in `org.eclipse.xtend.ide.tests/longrunning/src`, and so I would assume this is the launcher I'm supposed to use to kick off the tests. I was originally putting my tests next to `SpellingQuickfixTest`, but I think this was incorrect based on the package name shown in the example stub.

[...]

PR #1085, opened 2020-07-12
[eclipse/xtext-xtend#1061] Provide ui test infrastructure for testing (WIP)

drkstr101:
multi quickfixes [2/3]

Added `EqualsWithNullMultiQuickfixTest` and
`RemoveUnnecessaryModifiersMultiQuickfixTest`

Signed-off-by: Aaron R Miller <aaron.miller@waweb.io>

### DIFF
--- a/org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/RemoveUnnecessaryModifiersMultiQuickfixTest.xtend
+++ b/org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/RemoveUnnecessaryModifiersMultiQuickfixTest.xtend
@@ -1,0 +1,141 @@
+/*******************************************************************************
+ * Copyright (c) 2020 itemis AG (http://www.itemis.eu) and others.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.xtend.ide.tests.quickfix
+
+import org.eclipse.xtend.ide.tests.XtendIDEInjectorProvider
+import org.eclipse.xtext.testing.InjectWith
+import org.eclipse.xtext.testing.XtextRunner
+import org.eclipse.xtext.ui.editor.quickfix.QuickAssistCompletionProposal
+import org.eclipse.xtext.ui.testing.AbstractMultiQuickfixTest
+import org.junit.Test
+import org.junit.runner.RunWith
+
+import static extension org.eclipse.xtend.ide.tests.WorkbenchTestHelper.createPluginProject
+
+/**
+ * @author Aaron R Miller - Initial contribution and API
+ */
+@RunWith(XtextRunner)
+@InjectWith(XtendIDEInjectorProvider)
+class RemoveUnnecessaryModifiersMultiQuickfixTest extends AbstractMultiQuickfixTest {
+
+	static final String MODEL_WITH_UNNECESSARY_MODIFIERS = '''
+		package unnecessarymodifiers
+		public class Foo {
+			private final val A = 1
+			def public m() {
+				new Runnable {
+					public override def run() {
+						println(A)
+					}
+				}
+			}
+		}
+	'''
+
+	override String getFileName() {
+		return "Foo"
+	}
+
+	override dslFile(CharSequence content) {
+		super.dslFile(projectName, "src/unnecessarymodifiers/" + fileName, fileExtension, content);
+	}
+
+	override void setUp() throws Exception {
+		super.setUp()
+
+		projectName.createPluginProject()
+	}
+
+	@Test
+	def void testRemoveUnnecessaryModifiersQuickfixXtend() {
+
+		// Use Xtend style API...
+		
+		// TODO...
+		// org.junit.ComparisonFailure: expected:<...nnecessarymodifiers
+		// [class Foo {
+		// 	 val A = 1
+		//	 def m() {
+		//		 new Runnable {
+		//			 override run() {
+		//				 println(A)
+		//			 }
+		//		 }
+		//	 }
+		// }
+		// ------------------
+		// (no markers found)]> but was:<...nnecessarymodifiers
+		// [<0<public>0> class Foo {
+		//	 <1<private>1> <2<final>2> val A = 1
+		//	 def <3<public>3> m() {
+		//		 new Runnable {
+		//			 <4<public>4> override <5<def>5> run() {
+		//				 println(A)
+		//			 }
+		//		 }
+		//	 }
+		// }
+		// ---------------------------------------------------
+		// 0: message=The public modifier is unnecessary on class Foo
+		// 1: message=The private modifier is unnecessary on field A
+		// 2: message=The final modifier is unnecessary on field A
+		// 3: message=The public modifier is unnecessary on method m
+		// 4: message=The public modifier is unnecessary on method run
+		// 5: message=The def modifier is unnecessary on method run]
+		MODEL_WITH_UNNECESSARY_MODIFIERS.testMultiQuickfix('''
+			package unnecessarymodifiers
+			<0<public>0> class Foo {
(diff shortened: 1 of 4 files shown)


drkstr101:
TODO RemoveUnnecessaryModifiersMultiQuickfixTest contains failing test cases

org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/RemoveUnnecessaryModifiersMultiQuickfixTest.xtend
```diff
@@ -0,0 +1,120 @@
...
+				new Runnable {
+					public override def run() {
+						println(A)
+					}
+				}
+			}
+		}
+	'''
+
+	XtextEditor xtextEditor
```
miklossy, line 43: Move the `xtextEditior` variable into the `testRemoveUnnecessaryModifiersQuickfixJava` method

org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/RemoveUnnecessaryModifiersMultiQuickfixTest.xtend
```diff
@@ -0,0 +1,120 @@
...
+							println(A)
+						}
+					}
+				}
+			}
+			----------------------------------------------------
+			0: message=The public modifier is unnecessary on class Foo
+			1: message=The private modifier is unnecessary on field A
+			2: message=The final modifier is unnecessary on field A
+			3: message=The value of the field Foo.A is not used
```
miklossy, line 82: Remove `3: message=The value of the field Foo.A is not used` from the expected values.

org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/RemoveUnnecessaryModifiersMultiQuickfixTest.xtend
```diff
@@ -0,0 +1,120 @@
...
+
+	@Test
+	def void testRemoveUnnecessaryModifiersQuickfixJava() {
+
+		// Use Java style API
+		// //
+		
+		xtextEditor = openEditor(dslFile(MODEL_WITH_UNNECESSARY_MODIFIERS))
+
+		val offset = MODEL_WITH_UNNECESSARY_MODIFIERS.indexOf("==") + 1
```
miklossy, line 111: Replace `==` by `public`.

org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/RemoveUnnecessaryModifiersMultiQuickfixTest.xtend
```diff
@@ -0,0 +1,120 @@
...
+		val proposals = computeQuickAssistProposals(xtextEditor, offset)
+
+		// TODO java.lang.AssertionError: expected:<1> but was:<0>
+		assertEquals(1, proposals.size())
+		assertEquals(1, proposals.filter[it instanceof QuickAssistCompletionProposal].size())
+
+		// TODO assert proposal text...
```
miklossy, line 118: Add an assertion here, e.g. `assertEquals("Remove the unnecessary modifier.", proposals.head.displayString)`

miklossy:
> TODO RemoveUnnecessaryModifiersMultiQuickfixTest contains failing test cases

It is actually quite strange for me also why the same setup works for the `EqualsWithNullMultiQuickfixTest` but not for the `RemoveUnnecessaryModifiersMultiQuickfixTest`.

With the following workaround the `RemoveUnnecessaryModifiersMultiQuickfixTest` succeed as well:


import org.eclipse.core.resources.IFile
import org.eclipse.core.resources.IMarker
import org.eclipse.core.runtime.NullProgressMonitor
...

class RemoveUnnecessaryModifiersMultiQuickfixTest extends AbstractMultiQuickfixTest {
	...
	// TODO: verify why this workaround is necessary
	override protected IMarker[] getMarkers(IFile file) {
		saveAllOpenEditors
		super.getMarkers(file)
	}

	private def saveAllOpenEditors() {
		activePage.editorReferences.forEach[
			getEditor(false).doSave(new NullProgressMonitor)
		]
	}
}

drkstr101:
@miklossy Thanks for the input! I will make the changes you outlined in your comments and resubmit.

Cheers!

miklossy:
@drkstr101 How is the progress here?

drkstr101:
@miklossy I apologize, I meant to leave a comment earlier. I am having trouble getting the input strings just right in `RemoveUnnecessaryModifiersMultiQuickfixTest`. The markers seem to be correct, I believe my `---` separator may be the incorrect length. I am about to take another stab at it now.

org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/RemoveUnnecessaryModifiersMultiQuickfixTest.xtend
```diff
@@ -0,0 +1,141 @@
...
+
+		projectName.createPluginProject()
+	}
+
+	@Test
+	def void testRemoveUnnecessaryModifiersQuickfixXtend() {
+
+		// Use Xtend style API...
+		
+		// TODO...
```
drkstr101, line 61: @miklossy here is the error output

miklossy: @drkstr101 : In order to make the `RemoveUnnecessaryModifiersMultiQuickfixTest` test cases succeed, you have to ensure that the editors are saved before calculating the markers, see my previous [comment](https://github.com/eclipse/xtext-xtend/pull/1085#issuecomment-657361334) Could you please add these two methods (`getMarkers`, `saveAllOpenEditors`) to your PR?

miklossy:
@drkstr101 : Thanks for your contribution Aaron! I am merging this PR right now and will continue to complete this issue.

miklossy:
@drkstr101 I created a follow up PR: https://github.com/eclipse/xtext-xtend/pull/1090

drkstr101:
@miklossy @cdietrich 

Sorry you had to pick up a bit of the slack for me here. It looks like everything is taken care of now, but I still have a few notifications I need to touch. I do have some more free time this week if you have anything else you would like me to take a look at.

I also have some feedback on `AbstractMultiQuickfixTest` api if you are interested. My only comment is that I found it somewhat difficult to guess at the expected input format of the input strings. I had to do a lot of trial and error fiddling, and it never was fully clear what the last issue was since the important text was truncated in the output. It is also possible I was just missing something obvious in the assertion error. 

Coming up with a better solution is probably still a bit over my pay grade at this point however. I am starting to learn more of the internals though, if there was anything else you would like me to dig into. I can commit to roughly 30 hours or so over the course of this week(end).

miklossy:
@drkstr101 : Thanks for further offering your support! We think that https://github.com/eclipse/xtext-core/issues/897 would be a valuable contribution to the Xtext framework and we would be happy if you could take a look at that.

cdietrich:
@drkstr101 feedback for AbstractMultiQuickfixTest is also welcome

drkstr101:
@cdietrich My biggest complaint was requiring multiple runs to adjust my `-----`  line separators to get the desired length. Perhaps I need to adjust my font settings. This was amplified due to fiddling with the marker text which would then affect the line length and require more adjusting. I think it would be a fairly solid API if those line separators could be eliminated somehow.

szarnekow:
I agree completely. The separators are cumbersome and don't provide added value.

drkstr101:
Also worth noting, my time to implement the tests was effectively limited to the number of times I could launch the test workspace per hour, which was not very high on my machine unfortunately. This probably wouldn't be so bad if I wasn't depending on the test runs to discover what the inputs should be. `#badtester`

PR #1140, opened 2020-11-16
regenerated languages

cdietrich:
regenerated languages

### DIFF
--- a/org.eclipse.xtend.core/src-gen/org/eclipse/xtend/core/services/XtendGrammarAccess.java
+++ b/org.eclipse.xtend.core/src-gen/org/eclipse/xtend/core/services/XtendGrammarAccess.java
@@ -50,7 +50,9 @@ public class XtendGrammarAccess extends AbstractElementFinder.AbstractGrammarEle
 		//	xtendTypes+=Type*;
 		@Override public ParserRule getRule() { return rule; }
 		
-		//('package' package=QualifiedName ';'?)? importSection=XImportSection? xtendTypes+=Type*
+		//('package' package=QualifiedName ';'?)?
+		//importSection=XImportSection?
+		//xtendTypes+=Type*
 		public Group getGroup() { return cGroup; }
 		
 		//('package' package=QualifiedName ';'?)?
@@ -184,8 +186,8 @@ public class XtendGrammarAccess extends AbstractElementFinder.AbstractGrammarEle
 		//	members+=Member*
 		//	'}'
 		//	| {XtendInterface.annotationInfo=current} modifiers+=CommonModifier*
-		//	'interface' name=ValidID ('<' typeParameters+=JvmTypeParameter (',' typeParameters+=JvmTypeParameter)* '>')?
-		//	('extends' extends+=JvmSuperTypeReference (',' extends+=JvmSuperTypeReference)*)? '{'
+		//	'interface' name=ValidID ('<' typeParameters+=JvmTypeParameter (',' typeParameters+=JvmTypeParameter)* '>')? (
+		//	'extends' extends+=JvmSuperTypeReference (',' extends+=JvmSuperTypeReference)*)? '{'
 		//	members+=Member*
 		//	'}'
 		//	| {XtendEnum.annotationInfo=current} modifiers+=CommonModifier*
@@ -197,16 +199,24 @@ public class XtendGrammarAccess extends AbstractElementFinder.AbstractGrammarEle
 		//	'}');
 		@Override public ParserRule getRule() { return rule; }
 		
-		//{XtendTypeDeclaration} annotations+=XAnnotation* ({XtendClass.annotationInfo=current} modifiers+=CommonModifier* 'class'
-		//name=ValidID ('<' typeParameters+=JvmTypeParameter (',' typeParameters+=JvmTypeParameter)* '>')? ("extends"
+		//{XtendTypeDeclaration} annotations+=XAnnotation* ({XtendClass.annotationInfo=current} modifiers+=CommonModifier*
+		//'class' name=ValidID ('<' typeParameters+=JvmTypeParameter (',' typeParameters+=JvmTypeParameter)* '>')? ("extends"
 		//extends=JvmParameterizedTypeReference)? ('implements' implements+=JvmSuperTypeReference (','
-		//implements+=JvmSuperTypeReference)*)? '{' members+=Member* '}' | {XtendInterface.annotationInfo=current}
-		//modifiers+=CommonModifier* 'interface' name=ValidID ('<' typeParameters+=JvmTypeParameter (','
-		//typeParameters+=JvmTypeParameter)* '>')? ('extends' extends+=JvmSuperTypeReference (','
-		//extends+=JvmSuperTypeReference)*)? '{' members+=Member* '}' | {XtendEnum.annotationInfo=current}
-		//modifiers+=CommonModifier* 'enum' name=ValidID '{' (members+=XtendEnumLiteral (',' members+=XtendEnumLiteral)*)? ';'?
-		//'}' | {XtendAnnotationType.annotationInfo=current} modifiers+=CommonModifier* 'annotation' name=ValidID '{'
-		//members+=AnnotationField* '}')
+		//implements+=JvmSuperTypeReference)*)? '{'
+		//members+=Member*
+		//'}'
+		//| {XtendInterface.annotationInfo=current} modifiers+=CommonModifier*
+		//'interface' name=ValidID ('<' typeParameters+=JvmTypeParameter (',' typeParameters+=JvmTypeParameter)* '>')? ('extends'
+		//extends+=JvmSuperTypeReference (',' extends+=JvmSuperTypeReference)*)? '{'
+		//members+=Member*
+		//'}'
+		//| {XtendEnum.annotationInfo=current} modifiers+=CommonModifier*
+		//'enum' name=ValidID '{' (members+=XtendEnumLiteral (',' members+=XtendEnumLiteral)*)? ';'?
+		//'}'
+		//| {XtendAnnotationType.annotationInfo=current} modifiers+=CommonModifier*
+		//'annotation' name=ValidID '{'
+		//members+=AnnotationField*
+		//'}')
 		public Group getGroup() { return cGroup; }
 		
 		//{XtendTypeDeclaration}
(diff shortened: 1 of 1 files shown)


org.eclipse.xtend.core/src-gen/org/eclipse/xtend/core/services/XtendGrammarAccess.java
```diff
@@ -177,36 +179,44 @@
 		private final Keyword cRightCurlyBracketKeyword_2_3_6 = (Keyword)cGroup_2_3.eContents().get(6);
 		
 		//Type XtendTypeDeclaration:
-		//	{XtendTypeDeclaration} annotations+=XAnnotation* ({XtendClass.annotationInfo=current} modifiers+=CommonModifier*
+		//	{XtendTypeDeclaration} annotations+=XAnnotation* ({XtendClass . annotationInfo = current} modifiers+=CommonModifier*
```
kthoms, line 182: sorry for the question, but why did this change? It was better before IMHO.

cdietrich: please report this as bug to arne

cdietrich: https://github.com/eclipse/xtext-core/issues/1489#issuecomment-728713386

kthoms: OK, thanks for reporting. How about this change then? Would you discard the changes to XtendGrammarAccess then and wait for a fix? Or would you like to commit them now and revert with the next generation after the formatter has been fixed?

szarnekow: I'd like to get @ArneDeutsch feedback on the findings. Afterwards we can make an informed decision.

org.eclipse.xtend.core/src-gen/org/eclipse/xtend/core/services/XtendGrammarAccess.java
```diff
@@ -6589,10 +6727,10 @@ public ParserRule getXFunctionTypeRefRule() {
 	}
 	
 	//JvmParameterizedTypeReference:
-	//	type=[JvmType|super::QualifiedName] (=> '<' arguments+=JvmArgumentTypeReference (','
-	//	arguments+=JvmArgumentTypeReference)* '>' (=> ({JvmInnerTypeReference.outer=current} '.')
-	//	type=[JvmType|super::ValidID] (=> '<' arguments+=JvmArgumentTypeReference (',' arguments+=JvmArgumentTypeReference)*
-	//	'>')?)*)?;
+	//	type=[JvmType|super::QualifiedName] (=>'<' arguments+=JvmArgumentTypeReference (','
+	//	arguments+=JvmArgumentTypeReference)* '>' (=> ({JvmInnerTypeReference.outer=current} '.') type=
```
cdietrich, line 6731: more unexpected newlines. @szarnekow what do you think?

org.eclipse.xtend.core/src-gen/org/eclipse/xtend/core/services/XtendGrammarAccess.java
```diff
@@ -197,16 +199,24 @@
...
+		//'class' name=ValidID ('<' typeParameters+=JvmTypeParameter (',' typeParameters+=JvmTypeParameter)* '>')? ("extends"
 		//extends=JvmParameterizedTypeReference)? ('implements' implements+=JvmSuperTypeReference (','
-		//implements+=JvmSuperTypeReference)*)? '{' members+=Member* '}' | {XtendInterface.annotationInfo=current}
-		//modifiers+=CommonModifier* 'interface' name=ValidID ('<' typeParameters+=JvmTypeParameter (','
-		//typeParameters+=JvmTypeParameter)* '>')? ('extends' extends+=JvmSuperTypeReference (','
-		//extends+=JvmSuperTypeReference)*)? '{' members+=Member* '}' | {XtendEnum.annotationInfo=current}
-		//modifiers+=CommonModifier* 'enum' name=ValidID '{' (members+=XtendEnumLiteral (',' members+=XtendEnumLiteral)*)? ';'?
-		//'}' | {XtendAnnotationType.annotationInfo=current} modifiers+=CommonModifier* 'annotation' name=ValidID '{'
-		//members+=AnnotationField* '}')
+		//implements+=JvmSuperTypeReference)*)? '{'
```
szarnekow, line 205: Nice improvement over the old formatter @ArneDeutsch

org.eclipse.xtend.core/src-gen/org/eclipse/xtend/core/services/XtendGrammarAccess.java
```diff
@@ -329,9 +349,11 @@
 		//'}'
 		public Keyword getRightCurlyBracketKeyword_2_0_9() { return cRightCurlyBracketKeyword_2_0_9; }
 		
-		//{XtendInterface.annotationInfo=current} modifiers+=CommonModifier* 'interface' name=ValidID ('<'
-		//typeParameters+=JvmTypeParameter (',' typeParameters+=JvmTypeParameter)* '>')? ('extends'
-		//extends+=JvmSuperTypeReference (',' extends+=JvmSuperTypeReference)*)? '{' members+=Member* '}'
+		//{XtendInterface.annotationInfo=current} modifiers+=CommonModifier*
+		//'interface' name=ValidID ('<' typeParameters+=JvmTypeParameter (',' typeParameters+=JvmTypeParameter)* '>')? ('extends'
+		//extends+=JvmSuperTypeReference (',' extends+=JvmSuperTypeReference)*)? '{'
+		//members+=Member*
```
szarnekow, line 355: I'd really like to carry over indentation from the original source.

org.eclipse.xtend.core/src-gen/org/eclipse/xtend/core/services/XtendGrammarAccess.java
```diff
@@ -415,8 +437,9 @@
 		//'}'
 		public Keyword getRightCurlyBracketKeyword_2_1_8() { return cRightCurlyBracketKeyword_2_1_8; }
 		
-		//{XtendEnum.annotationInfo=current} modifiers+=CommonModifier* 'enum' name=ValidID '{' (members+=XtendEnumLiteral (','
-		//members+=XtendEnumLiteral)*)? ';'? '}'
+		//{XtendEnum.annotationInfo=current} modifiers+=CommonModifier*
+		//'enum' name=ValidID '{' (members+=XtendEnumLiteral (',' members+=XtendEnumLiteral)*)? ';'?
```
szarnekow, line 441: The original source has more line breaks. If the goal is to preserve line-breaks, there is something to do here.

szarnekow (approved):

org.eclipse.xtend.core/src-gen/org/eclipse/xtend/core/services/XtendGrammarAccess.java
ArneDeutsch: Fixed with next commit. Missing recursion.

org.eclipse.xtend.core/src-gen/org/eclipse/xtend/core/services/XtendGrammarAccess.java
ArneDeutsch: After the next patch this
```
ParameterizedTypeReferenceWithTypeArgs returns types::JvmParameterizedTypeReference:
	type=[types::JvmType|QualifiedName] (
		'<' arguments+=JvmArgumentTypeReference (',' arguments+=JvmArgumentTypeReference)* '>'
		(=>({types::JvmInnerTypeReference.outer=current} '.') type=[types::JvmType|ValidID] (=>'<' arguments+=JvmArgumentTypeReference (',' arguments+=JvmArgumentTypeReference)* '>')?)*
	)
;
```
Will be formatted as
```
ParameterizedTypeReferenceWithTypeArgs returns types::JvmParameterizedTypeReference:
	type=[types::JvmType|QualifiedName] ('<' arguments+=JvmArgumentTypeReference (','
	arguments+=JvmArgumentTypeReference)* '>'
	(=>({types::JvmInnerTypeReference.outer=current} '.') type=[types::JvmType|ValidID] (=>'<'
	arguments+=JvmArgumentTypeReference (',' arguments+=JvmArgumentTypeReference)* '>')?)*);
```
Looks ok to me. The remaining newlines are the result of autowrap because otherwise the lines get to long.

szarnekow: Curiosity: Did you try with indent between the opening and closing parentheses?

szarnekow: Also: If we keep the newlines from the original source, why is it removed after the opening `)` in the first line?

org.eclipse.xtend.core/src-gen/org/eclipse/xtend/core/services/XtendGrammarAccess.java
ArneDeutsch: Right now we preserve NOT ALL linebreaks but only the ones between tokens. To allow something like:
```
A:
  B
  C
  D;
```
or 
```
A:
  b=B
  c=C
  d=D;
```
But we do not preserve linebreaks to other positions (e.g. after parantesis or braces). In result:
```
A:
  (
  b=B
  c=C
  d=D
  );
```
will become
```
A:
  (b=B
  c=C
  d=D);
```

szarnekow: My gut feeling (without having done research on the existing sample grammars in our code base): Especially the manually inserted line breaks in the last example should be kept. At least I *think* that I often put line breaks (and indentation) in parenthesised groups to make the grammar more readable. What do you think?

ArneDeutsch: Well, I think it is hard to decide ;). Of course many of the formattings are intentionally and it would be good to preserve them. On the other hand ... why do we have a formatter then? If we decide to do not touch the formatting we would have a null-Formatter. So in the end it is a compromise anyway.

Ritght now we say, that a well formatted grammer has every rule in one line (if it fits, otherwise we break at sensible points). But we support a bit of custom formatting to allow formatting like the following:
```
ClassDecl:
    'class' name=ID ('extends' superClass=ClassDecl)? '{'
    body=ClassBody
    '}';
```
Of course badly formatted stuff like the following would benefit from a more strict formatter (here we see the compromise):
```
ClassDecl:
    'class'
    name=ID ('extends'
    superClass=ClassDecl)? '{'
    body=ClassBody '}';
```
But at least our formatter can fix the following glitch:
```
ClassDecl:
    'class' name=ID ('extends' superClass=ClassDecl
    )? '{'
    body=ClassBody
    '}';
```
If we made the formatter even more tollerant to custom formatting, he would not be able any more to move the parenthesis one line above.

Because of this reasoning I would argue for a more strict formatting in most cases.

org.eclipse.xtend.core/src-gen/org/eclipse/xtend/core/services/XtendGrammarAccess.java
ArneDeutsch: Right now (and in the past) we maintain newlines only between tokens, not before or after seperators. For a discussion see above.

org.eclipse.xtend.core/src-gen/org/eclipse/xtend/core/services/XtendGrammarAccess.java
szarnekow: We should be careful to not confuse the different use cases for the formatter:
- Editor
- Serialization from scratch
- Grammar access (serialization with node model)

For the grammar access, it may be ok to use a null-formatting. For the other cases, we still need a formatter.

Issue #1143, opened 2020-11-23
Xtend Activator Problem


miklossy:
After updating to Xtend 2.24.0.v20201123-0220, I observe a problem with the Xtend Activator:

![screencast](https://user-images.githubusercontent.com/20393472/99933206-a3012580-2d5a-11eb-94be-2be1d1318157.gif)

**Environment:**
Eclipse IDE 2020-06 (4.16.0)
Eclipse Xtext 2.24.0.v20201123-0220
Eclipse Xtend 2.24.0.v20201123-0220

**Hint:**
It was working till Xtend 2.24.0.v20201120-0220, so the problem could be related to the [last change](https://github.com/eclipse/xtext-xtend/commit/45aa82b252859db51c3e1bb3cc2f3ea45650bbc0) on the Xtend Activator.

@cdietrich @kthoms Can you reproduce the problem on your computer?

cdietrich:
Please do start the bundle via Osgi console and Post error Message

cdietrich:
and no i cant. but i head problems with xtend classpath container yesterday which somehow resolved magically by restarting

miklossy:
I also observe some problem with the Xtend classpaht container. Unfortunately, they does not disapper by restarting:
![screenshot](https://user-images.githubusercontent.com/20393472/99936908-98e42480-2d64-11eb-9914-69fd70cc9672.png)

miklossy:
> 
> 
> Please do start the bundle via Osgi console and Post error Message
```

osgi> start 2897
org.osgi.framework.BundleException: Error loading bundle activator. org.eclipse.xtend.ide_2.24.0.v20201122-0706 [2897]
	at org.eclipse.osgi.internal.framework.BundleContextImpl.start(BundleContextImpl.java:786)
	at org.eclipse.osgi.internal.framework.EquinoxBundle.startWorker0(EquinoxBundle.java:1015)
	at org.eclipse.osgi.internal.framework.EquinoxBundle$EquinoxModule.startWorker(EquinoxBundle.java:365)
	at org.eclipse.osgi.container.Module.doStart(Module.java:605)
	at org.eclipse.osgi.container.Module.start(Module.java:468)
	at org.eclipse.osgi.internal.framework.EquinoxBundle.start(EquinoxBundle.java:439)
	at org.apache.felix.gogo.command.Basic.start(Basic.java:739)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #1144, opened 2020-11-23
[#1143] Disable theme manager due to regression

kthoms:
It was enough to disable the event listener. No need to change the plugin.xml to come back to the previous state.

Bildschirmfoto 2020-11-23 um 09 34 13
Bildschirmfoto 2020-11-23 um 09 34 29

The dark theme support will be reconsidered for 2.25.

### DIFF
--- a/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/XtendActivator.java
+++ b/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/XtendActivator.java
@@ -8,29 +8,19 @@
  *******************************************************************************/
 package org.eclipse.xtend.ide;
 
-import static java.util.Collections.*;
-
-import java.util.Optional;
-
-import org.eclipse.e4.core.services.events.IEventBroker;
-import org.eclipse.e4.ui.css.swt.theme.IThemeEngine;
-import org.eclipse.swt.widgets.Display;
-import org.eclipse.ui.PlatformUI;
-import org.eclipse.xtend.ide.highlighting.XtendThemeManager;
 import org.osgi.framework.BundleContext;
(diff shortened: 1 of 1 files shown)


org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/XtendActivator.java
```diff
@@ -8,36 +8,32 @@
...
  */
-@SuppressWarnings("restriction")
 public class XtendActivator extends org.eclipse.xtend.ide.internal.XtendActivator {
 
 	private XtendThemeManager themeManager;
 
 	@Override
 	public void start(BundleContext context) throws Exception {
 		super.start(context);
+		/* disabled due to issue#1173
```
cdietrich, line 28: what about stop & plugin.xml? change not needed?

kthoms: not necessary. themeManager will be null, which is guarded in stop()
plugin.xml has no negative effect

kthoms: if you want I can comment out the code in stop()

szarnekow (approved):

szarnekow:
Code looks good. I didn't try it to see if it fixes the regression.

org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/XtendActivator.java
kthoms: done

cdietrich (approved):

Issue #730, opened 2019-02-20
Better code generation for dispatch functions with Void and Object


tivervac:
When you use a dispatch with a `Void` argument, the generated code checks if the argument is `null`.

When you use a dispatch with an `Object` argument, the generated code checks if the argument is not `null`.

If you use them as a combination both will happen, however it will always be known that the argument is not null.

For example, the following Xtend:
```
package com.test.pkg

class Test {
	def dispatch void doThing(Void v) {}
	def dispatch void doThing(Object o) {}
}
```

Generates the following Java:
```
package com.test.pkg;

import java.util.Arrays;

@SuppressWarnings("all")
public class Test {
  protected void _doThing(final Void v) {
  }
  
  protected void _doThing(final Object o) {
  }
  
  public void doThing(final Object v) {
    if (v == null) {
      _doThing((Void)null);
      return;
    // v is known to be not null here making the else if and else useless, there should just be an else
    } else if (v != null) {
      _doThing(v);
      return;
    } else {
      throw new IllegalArgumentException("Unhandled parameter types: " +
        Arrays.<Object>asList(v).toString());
    }
  }
}
```

miklossy:
See also https://github.com/eclipse/xtext-xtend/issues/174

ArneDeutsch:
Investigated this for quite some time without coming to a working solution. A fix would be needed in `DispatchMethodCompileStrategy`.
Issues that make the issue complicated:
- One `if` is generated for every method. Exact content of each `if` statement is calculated for each one seperately. One would need to go through this list of methods one extra round and compute if generation is necessary in each case. This would probably make the generation process much slower. Because type information is needed for the process.
- Case is simple only for methods with exactly one parameter. It exists for multiple parameters as well and would be very hard to solve in general case. Example:

```
class Test {
	def dispatch void doThing(Object v, Object v1) {}
	def dispatch void doThing(Void v, Object v1) {}
	def dispatch void doThing(Object v, Void v1) {}
	def dispatch void doThing(Void v, Void v1) {}
}
```
```
  public void doThing(final Object v, final Object v1) {
    if (v == null
         && v1 == null) {
      _doThing((Void)null, (Void)null);
      return;
    } else if (v == null
         && v1 != null) {
      _doThing((Void)null, v1);
      return;
    } else if (v != null
         && v1 == null) {
      _doThing(v, (Void)null);
      return;
    } else if (v != null
         && v1 != null) {
      _doThing(v, v1);
      return;
[...]

PR #1179, opened 2021-02-26
[#730] Skip generation of else and last case if all cases are covered.

ArneDeutsch:


ArneDeutsch:
When reviewing plase have an eye on the following questions:
# Is there a risk that performance suffers? I have used some method calls in the first loop from which I can not judge well the performance costs.
# Is the missing return statement for the last case an issue (when the else is missing). Or good as it is?
# Do you have better ideas for the names "voidIncluded" and "notNullIncluded"? They are a bit technical, but I do not understand it well enough to find better names.

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/DispatchCompilerTest.xtend
```diff
@@ -497,5 +497,120 @@ class DispatchCompilerTest extends AbstractXtendCompilerTest {
...
+			  }
+			  
+			  protected void _doThing(final Object o) {
+			  }
+			  
+			  public void doThing(final Object v) {
+			    if (v == null) {
+			      _doThing((Void)null);
+			      return;
+			    } else _doThing(v);
```
szarnekow, line 521: Nit: Can you wrap the else statement in a block?

org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
```diff
@@ -58,23 +58,53 @@ protected void initialize(JvmOperation dispatchOperation) {
...
+				JvmTypeReference caseParameterType = caseParam.getParameterType();
+				LightweightTypeReference caseParamType = owner.toLightweightTypeReference(caseParam.getParameterType());
 				if (!Strings.equal(dispatchParameterType.getIdentifier(), caseParameterType.getIdentifier())) {
 					allCasesSameType[i] = false;
-					break;
+				}
+				if (caseParamType.isType(Void.class)) {
+					voidIncluded[i] = true;
+				}
+				TypeConformanceComputationArgument rawNoSynonyms = new TypeConformanceComputationArgument(true, false, true, true, false, false);
```
szarnekow, line 86: You might want to use `LightweightTypeReference.getIdentifier()` similar to what is done in `XbaseValidator.checkCast(JvmTypeReference, LightweightTypeReference, LightweightTypeReference)` in the last condition.
But I see that this is basically reused code from below. Can you please extract a private method for that? Something like `isSameType(..)` or `needsTypeGuard(..)`?

org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
```diff
@@ -58,23 +58,53 @@ protected void initialize(JvmOperation dispatchOperation) {
...
 		boolean[] allCasesSameType = new boolean[parameterCount];
+		boolean[] voidIncluded = new boolean[parameterCount];
+		boolean[] notNullIncluded = new boolean[parameterCount];
 		for(int i = 0; i < parameterCount; i++) {
 			allCasesSameType[i] = true;
-			JvmTypeReference dispatchParameterType = dispatchOperation.getParameters().get(i).getParameterType();
+			voidIncluded[i] = false;
+			notNullIncluded[i] = false;
+			JvmFormalParameter dispatchParam = dispatchOperation.getParameters().get(i);
+			LightweightTypeReference dispatchParamType = owner.toLightweightTypeReference(dispatchParam.getParameterType());
```
szarnekow, line 74: I find these names hard to distinguish - dispatchParamType vs dispatchParameterType

org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
```diff
@@ -58,23 +58,53 @@ protected void initialize(JvmOperation dispatchOperation) {
...
+			JvmTypeReference dispatchParameterType = dispatchParam.getParameterType();
 			for (JvmOperation operation : sortedDispatchOperations) {
-				JvmFormalParameter parameter = operation.getParameters().get(i);
-				JvmTypeReference caseParameterType = parameter.getParameterType();
+				JvmFormalParameter caseParam = operation.getParameters().get(i);
+				JvmTypeReference caseParameterType = caseParam.getParameterType();
+				LightweightTypeReference caseParamType = owner.toLightweightTypeReference(caseParam.getParameterType());
 				if (!Strings.equal(dispatchParameterType.getIdentifier(), caseParameterType.getIdentifier())) {
```
szarnekow, line 80: It should be possible to use getIdentifier from the LightweightTypeReference here and thereby avoid the need for the JvmTypeReference local vars.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
```diff
@@ -58,23 +58,53 @@ protected void initialize(JvmOperation dispatchOperation) {
...
 					allCasesSameType[i] = false;
-					break;
+				}
+				if (caseParamType.isType(Void.class)) {
+					voidIncluded[i] = true;
```
szarnekow, line 84: Note to self: `voidIncluded` implies that there was a null check against the parameter. `notNullIncluded` implies that that the exact type was used.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
```diff
@@ -58,23 +58,53 @@ protected void initialize(JvmOperation dispatchOperation) {
...
+		}
+		if (parameterCount == 1 && sortedDispatchOperations.size() == 1)
+			needsElse = false;
+		for (boolean n : notNullIncluded) {
+			if (!n) {
+				needsLastCase = true;
+				break;
+			}
+		}
+		if (needsElse || (parameterCount == 1 && sortedDispatchOperations.size() == 1))
```
szarnekow, line 106: Line 99 sets `needsElse` to false if 
`parameterCount == 1 && sortedDispatchOperations.size() == 1` - here we have the same condition - needsElse is already provable false in that case. We can probably rewrite this to:
```
if (parameterCount == 1 && sortedDispatchOperations.size() == 1) {
  needsElse = false; needsLastCase = true;
} else {
  needsElse = !allTrue(voidIncluded);
  needsLastCase = !allTrue(notNullIncluded);
}
```

org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
```diff
@@ -58,23 +58,53 @@ protected void initialize(JvmOperation dispatchOperation) {
...
-		ITypeReferenceOwner owner = new StandardTypeReferenceOwner(services, dispatchOperation);
+		for (boolean v : voidIncluded) {
+			if (!v) {
+				needsElse = true;
+				break;
+			}
+		}
+		if (parameterCount == 1 && sortedDispatchOperations.size() == 1)
+			needsElse = false;
+		for (boolean n : notNullIncluded) {
```
szarnekow, line 100: Please extract a tiny method such that we can write
`needsElse = !allTrue(voidIncluded)` and `needsLastCase = !allTrue(notNullIncluded)`.
Please try to find a better name for `allTrue(boolean[] booleans)`

org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
```diff
@@ -58,23 +58,53 @@ protected void initialize(JvmOperation dispatchOperation) {
 	public void apply(/* @Nullable */ ITreeAppendable a) {
 		if (a == null)
 			throw new IllegalArgumentException("a is never null");
-		boolean needsElse = true;
+		boolean needsElse = false;
```
szarnekow, line 61: Please move the declarations of these locals closer to their first usage

szarnekow:
@ArneDeutsch  The approach looks fair enough to me. The comments are all only related to implementation details. I'm aware that the code was already messy as is. Nevertheless I'd greatly appreciate if you could extract a few smaller methods as outlined.

szarnekow:
> Is the missing return statement for the last case an issue (when the else is missing). Or good as it is?

I didn't spot that immediately. Can you elaborate? Which return is missing?

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/DispatchCompilerTest.xtend
szarnekow, line 521: You mean this location where there is no return?

ArneDeutsch: Yes, that is what I mean.

szarnekow: Perfectly fine for me.

ArneDeutsch: Will do.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
ArneDeutsch: Will do.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
ArneDeutsch: Me too. But to be honest ... I do not understand the code at this place at all, that's why it is hard for me to make good names here.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
ArneDeutsch: Yes, will change that. Will also solve the naming issue.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
ArneDeutsch: Will do.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
ArneDeutsch: Will do, althought a no much better name comes to mind.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
ArneDeutsch: Code does not work, leads to failing tests. But will simplify it a bit anyway.

szarnekow (approved):

PR #1183, opened 2021-03-05
[eclipse/xtext#1875] Remove redundant project version

kthoms:
The version is already inherited. The 'prepare-release-branches' only
replaces the first version occurance. The redundant second entry has to
be removed.

See https://github.com/eclipse/xtext-umbrella/pull/252#issuecomment-791181220


cdietrich (approved):
approved unter the ssumtion that the maven versions plugin is no longer used.

Issue #174, opened 2017-03-10
Xtend compiler generates dead code when dispatching with null value


miklossy:
![deadcode](https://cloud.githubusercontent.com/assets/20393472/23785568/a2c683da-0568-11e7-93d7-ef7d822c48b6.png)

cdietrich:
similar to https://bugs.eclipse.org/bugs/show_bug.cgi?id=481990
maybe dups https://bugs.eclipse.org/bugs/show_bug.cgi?id=393024

ArneDeutsch:
Done with https://github.com/eclipse/xtext-xtend/issues/730

cdietrich:
@ArneDeutsch  i now get errors on AbstractLanguageServerTest in xtend

```
protected String toExpectation(final Object it) {
    if (it instanceof Integer) {
      return _toExpectation((Integer)it);
    } else if (it instanceof List) {
      return _toExpectation((List<?>)it);
    } else if (it instanceof DocumentHighlightKind) {
      return _toExpectation((DocumentHighlightKind)it);
    } else if (it instanceof String) {
      return _toExpectation((String)it);
    } else if (it instanceof VersionedTextDocumentIdentifier) {
      return _toExpectation((VersionedTextDocumentIdentifier)it);
    } else if (it instanceof Pair) {
      return _toExpectation((Pair<SemanticHighlightingInformation, List<List<String>>>)it);
    } else if (it == null) {
      return _toExpectation((Void)null);
    } else if (it instanceof Map) {
      return _toExpectation((Map<Object, Object>)it);
    } else if (it instanceof CodeAction) {
      return _toExpectation((CodeAction)it);
    } else if (it instanceof CodeLens) {
      return _toExpectation((CodeLens)it);
    } else if (it instanceof Command) {
      return _toExpectation((Command)it);
    } else if (it instanceof CompletionItem) {
      return _toExpectation((CompletionItem)it);
    } else if (it instanceof DocumentHighlight) {
      return _toExpectation((DocumentHighlight)it);
    } else if (it instanceof DocumentSymbol) {
      return _toExpectation((DocumentSymbol)it);
    } else if (it instanceof Hover) {
      return _toExpectation((Hover)it);
    } else if (it instanceof Location) {
      return _toExpectation((Location)it);
    } else if (it instanceof MarkupContent) {
      return _toExpectation((MarkupContent)it);
    } else if (it instanceof Position) {
      return _toExpectation((Position)it);
[...]

PR #1194, opened 2021-04-03
[#174] Fix void case where not all cases are covered.

ArneDeutsch:

### DIFF
--- a/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/DispatchCompilerTest.xtend
+++ b/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/DispatchCompilerTest.xtend
@@ -9,7 +9,6 @@
 package org.eclipse.xtend.core.tests.compiler
 
 import org.junit.Test
-import org.junit.Ignore
 
 /**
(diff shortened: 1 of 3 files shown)


org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/DispatchCompilerTest.xtend
```diff
@@ -499,7 +498,7 @@ class DispatchCompilerTest extends AbstractXtendCompilerTest {
 		''')
 	}
 
-	@Test @Ignore
+	@Test
```
szarnekow, line 501: These ignored tests must have slipped through the cracks with the last review. Sorry about hat.

szarnekow (approved):

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/DispatchCompilerTest.xtend
cdietrich: no, i unrolled the last patch, thus had to ignore @ArneDeutsch  s new tests

PR #1216, opened 2021-08-05
use guice 5.0.1

cdietrich:
use guice 5.0.1


org.eclipse.xtend.core.tests.java8/META-INF/MANIFEST.MF
```diff
@@ -17,7 +17,7 @@ Require-Bundle: org.eclipse.xtend.core,
  org.hamcrest.core;bundle-version="1.3.0",
  org.eclipse.emf.codegen.ecore;bundle-version="2.20.0",
  com.google.guava;bundle-version="[30.1.0,31.0.0)",
- com.google.inject;bundle-version="3.0.0",
+ com.google.inject;bundle-version="5.0.1",
```
szarnekow, line 20: Should we generally use an upper bound exluding 5.1 or 6.0 ?
Do we trust semantic versioning?

org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/RemoveUnnecessaryModifiersMultiQuickfixTest.xtend
```diff
@@ -21,6 +21,7 @@ import org.junit.Test
 import org.junit.runner.RunWith
 
 import static extension org.eclipse.xtend.ide.tests.WorkbenchTestHelper.createPluginProject
+import org.eclipse.xtext.ui.testing.util.TargetPlatformUtil
```
szarnekow, line 24: Interesting

cdietrich: no idea why i do that

org.eclipse.xtend.core.tests.java8/META-INF/MANIFEST.MF
cdietrich: i have no idea we seem to use this pattern at multiple places so far.

PR #1217, opened 2021-08-12
update to gradle 7

cdietrich:
update to gradle 7
see https://github.com/eclipse/xtext/issues/1548


cdietrich:
i still see problem with the double acting
of core.tests as test and lib project
somehow stuff is not visible to java8 tests in eclipse

cdietrich:
Screenshot 2021-08-12 142822

cdietrich:
so the question is:
how can i have core.tests as api and tests. (reexport the tests)

org.eclipse.xtend.core.tests.java8/build.gradle
```diff
@@ -1,5 +1,6 @@
 dependencies {
-	compile project(':org.eclipse.xtend.core.tests')
+	testImplementation project(':org.eclipse.xtend.core.tests')
+	testImplementation project(':org.eclipse.xtend.core.tests').sourceSets.test.output // TODO is there a better way
```
oehme, line 3: Use the [test-fixtures](https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures) plugin to share the test fixtures from one project with another.

cdietrich: seems to work. thanks.

cdietrich: hmm not quite yet. will create follow up issue

szarnekow (approved):

PR #1321, opened 2022-04-11
fix xtend examples

cdietrich:
fix xtend examples


cdietrich:
@LorenzoBettini  looks like this one was missed.
we dont execute examples outside of
https://github.com/itemis/xtext-reference-projects/actions

PR #1331, opened 2022-04-24
[eclipse/xtext-eclipse#1842] remove test failure ignore

cdietrich:
[eclipse/xtext-eclipse#1842] remove test failure ignore


szarnekow (approved):
LGTM. Do we want to fail at the end or immediately? Probably immediately.

PR #1339, opened 2022-05-05
fixed some deprecations

cdietrich:
fixed some deprecations


szarnekow (approved):

cdietrich:
looks like the docs is improper. does not work with oxygen

Issue #1347, opened 2022-06-01
Update to new SWTBot 4.0.0


cdietrich:
Update to new SWTBot 4.0.0 (will be part of 2022-06)

cdietrich:
ommph already seems to use https://download.eclipse.org/technology/swtbot/releases/latest/

cdietrich:
unclear: how to solve logging problem
https://www.eclipse.org/lists/cross-project-issues-dev/msg19233.html

PR #1354, opened 2022-06-20
[#1347] update to swtbot 4.0.0

cdietrich:
[#1347] update to swtbot 4.0.0


org.eclipse.xtend.ide.swtbot.tests/pom.xml
```diff
@@ -54,4 +54,60 @@
 		</plugins>
   	</build>
 
+  	<profiles>
+		<profile>
+			<id>latest</id>
+			<build>
+				<plugins>
+					<plugin>
```
cdietrich, line 62: @kthoms is there a way to apply this to multiple profile ids without copy&pasting it over and over again

szarnekow: Maybe `<extraRequirements combine.children="append">` and only add the slf4j bundle?

cdietrich: does not seem to work 
```
[INFO] {osgi.os=win32, osgi.ws=win32, org.eclipse.update.install.features=true, osgi.arch=x86_64}
[ERROR] Cannot resolve project dependencies:
[ERROR]   You requested to install 'osgi.bundle; org.eclipse.jdt.launching.macosx 1.0.0' but it could not be found
```

cdietrich: maybe i can simply add it to all tests
need to check

PR #1371, opened 2022-08-14
Update to M2E 2.0

HannesWell:
This PR aims to update `org.eclipse.xtend.m2e` Plugin to the M2E 2.0.1 release that is contributed to Eclipse 2022-09 M3 SimRel.
M2E 2.0 is a major release that has breaking changes in its API, that are incompatible with previous 1.x releases.

I updated the code and adjusted required bundle version ranges accordingly. Furthermore I update the target file to include the latest m2e release 2.0.1. That release is not yet in the 2022-09 release repo but will be with M3, so this change in advance should prevent trouble with the xText contribution.
I hope that is all correct, but please let me know if anything is wrong or missing.

### DIFF
--- a/org.eclipse.xtend.m2e/src/org/eclipse/xtend/m2e/XtendProjectConfigurator.java
+++ b/org.eclipse.xtend.m2e/src/org/eclipse/xtend/m2e/XtendProjectConfigurator.java
@@ -15,9 +15,12 @@ import static org.eclipse.xtext.builder.preferences.BuilderPreferenceAccess.getK
 import static org.eclipse.xtext.builder.preferences.BuilderPreferenceAccess.getOutputForSourceFolderKey;
 
 import java.io.File;
+import java.lang.reflect.Method;
 import java.util.List;
+import java.util.function.Function;
 
 import org.apache.maven.plugin.MojoExecution;
+import org.apache.maven.project.MavenProject;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.ProjectScope;
 import org.eclipse.core.runtime.CoreException;
@@ -41,12 +44,10 @@ import com.google.common.base.Strings;
 public class XtendProjectConfigurator extends AbstractProjectConfigurator {
 
 	@Override
-	public void configure(ProjectConfigurationRequest request,
-			IProgressMonitor monitor) throws CoreException {
-		addNature(request.getProject(), XtextProjectHelper.NATURE_ID, monitor);
+	public void configure(ProjectConfigurationRequest request, IProgressMonitor monitor) throws CoreException {
+		addNature(getProject(request), XtextProjectHelper.NATURE_ID, monitor);
 
-		OutputConfiguration config = new XtendOutputConfigurationProvider()
-				.getOutputConfigurations().iterator().next();
+		OutputConfiguration config = new XtendOutputConfigurationProvider().getOutputConfigurations().iterator().next();
 
 		List<MojoExecution> executions = getMojoExecutions(request, monitor);
 		SubMonitor progress = SubMonitor.convert(monitor, executions.size());
@@ -63,7 +64,7 @@ public class XtendProjectConfigurator extends AbstractProjectConfigurator {
 			}
 		}
 
-		writePreferences(config, request.getProject());
+		writePreferences(config, getProject(request));
 	}
 
 	private void writePreferences(OutputConfiguration configuration,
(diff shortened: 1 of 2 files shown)


cdietrich:
I assume there is no way to make this runnable on oxygen .. latest and java 8 and java 11, at least without reflection

cdietrich:
I assume there is no way to make this runnable on oxygen .. latest and java 8 and java 11, at least without reflection

HannesWell:
> I assume there is no way to make this runnable on oxygen .. latest and java 8 and java 11, at least without reflection

I don't think so. M2E 2.0 even requires Java-17. If you want to use `org.eclipse.xtext.m2e` and `org.eclipse.xtend.m2e` with older Eclipse and M2E versions reflection has to be used. But since the changes here are relatively limited using reflection as fall back should be manageable.
The same applies to https://github.com/eclipse/xtext-eclipse/pull/1893.

Do you want me to change this PR in that way? If yes what should it be compiled against? M2E 2.0 or M2E 1.?

cdietrich:
as our default target platform still is oxygen and we also compile against phonton..2022-06 i would prefer to compile against 1x and use reflection to call 2x methods.
a PR for that would be highly welcome

HannesWell:
> as our default target platform still is oxygen and we also compile against photon..2022-06 i would prefer to compile against 1x and use reflection to call 2x methods. a PR for that would be highly welcome

Roger that. I'll update this PR and https://github.com/eclipse/xtext-eclipse/pull/1893 accordingly.


> 
> does the minimum j17 mean m2e cant be used on 2022-09 unless eclipse is run with J17? i am not sure how to test this. our builds against latest run with j11 and j17

Yes, exactly.
I also don't have an immediate answer to this. The xText build and project structure seem to be quite complex and I'm also not familiar with Gradle (only with Maven). But I would guess that only the J17 variant can use M2E then.

cdietrich:
the gradle part should not play a role. the tycho build can be run with
`-Poxygen` to run with oxygen, `-Platest` to run with latest and `-Pr202206` to run against 2022-06.
the tycho root pom is `-f tycho-pom.xml`

org.eclipse.xtend.m2e/src/org/eclipse/xtend/m2e/XtendProjectConfigurator.java
```diff
@@ -150,10 +151,33 @@ private void readTestDebugInfoConfig(OutputConfiguration config, ProjectConfigur
...
 	private <T> T mojoParameterValue(String paramName, Class<T> paramType, ProjectConfigurationRequest request,
 			MojoExecution execution, SubMonitor progress) throws CoreException {
-		return maven.getMojoParameterValue(request.getMavenProject(), execution, paramName, paramType,
-				progress.split(1));
+		return maven.getMojoParameterValue(getMavenProject(request), execution, paramName, paramType, progress.split(1));
+	}
+
+	static IProject getProject(ProjectConfigurationRequest request) {
+		// DO NOT USE A METHOD REFERENCE!
+		return call(request, r -> r.getMavenProjectFacade(), "mavenProjectFacade").getProject();
```
HannesWell, line 162: It would be good to have that tested.

cdietrich: Have started builds with 11 and 17 against latest too. Will try to have a look tomorrow or on Tuesday/Wednesday depending on how i find time

HannesWell:
> the gradle part should not play a role. the tycho build can be run with `-Poxygen` to run with oxygen, `-Platest` to run with latest and `-Pr202206` to run against 2022-06. the tycho root pom is `-f tycho-pom.xml`

At the moment I cannot think of a better way to have another `latest-j17` with dedicated target file where m2e 2.0 is used.

org.eclipse.xtend.m2e/src/org/eclipse/xtend/m2e/XtendProjectConfigurator.java
HannesWell: Thanks!

cdietrich:
manual test looks good. will check if i need todo something about j17 tp once M3 is out.

org.eclipse.xtend.m2e/src/org/eclipse/xtend/m2e/XtendProjectConfigurator.java
szarnekow: I'm afraid we have to compile against one version and run against the other to really test it.

cdietrich: i fear this is not tested at all. thus i created [Repo with patch](https://ci.eclipse.org/xtext/job/xtext-umbrella/job/cd_m2e20/lastSuccessfulBuild/artifact/build/p2-repository/) and did manual test (with remote debugging)

cdietrich:
thx @HannesWell

HannesWell:
Thank you and you are very welcome.

org.eclipse.xtend.m2e/src/org/eclipse/xtend/m2e/XtendProjectConfigurator.java
HannesWell: > I'm afraid we have to compile against one version and run against the other to really test it.

Yes. I cannot think of another way too.

But once M2E 2.0 is in SimRel Repo compiling against latest will also fail because then XText has to be compiled against the new version. So probably reflection has to be used for both versions until M2E 2.0 is the baseline for all?

PR #1372, opened 2022-08-15
m2e: use reflection also for old method

cdietrich:
m2e: use reflection also for old method

### DIFF
--- a/org.eclipse.xtend.m2e/src/org/eclipse/xtend/m2e/XtendProjectConfigurator.java
+++ b/org.eclipse.xtend.m2e/src/org/eclipse/xtend/m2e/XtendProjectConfigurator.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2013, 2018 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2013, 2022 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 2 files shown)


org.eclipse.xtend.m2e/src/org/eclipse/xtend/m2e/XtendProjectConfigurator.java
```diff
@@ -158,20 +158,19 @@ private <T> T mojoParameterValue(String paramName, Class<T> paramType, ProjectCo
...
 
 	static MavenProject getMavenProject(ProjectConfigurationRequest request) {
-		// DO NOT USE A METHOD REFERENCE!
-		return call(request, r -> r.getMavenProject(), "mavenProject");
+		return call(request, "getMavenProject", "mavenProject");
 	}
 
 	@SuppressWarnings("unchecked")
-	private static <T, R> R call(T obj, Function<T, R> getter, String newMethodName) {
+	private static <T, R> R call(T obj, String oldMethodName, String newMethodName) {
```
HannesWell, line 169: Alternatively we could use a varargs `methodNames` String arguments:
```suggestion
	private static <T, R> R call(T obj, String... methodNames) {
```
and then just loop over them and try each and throw an exception if none worked:
```
for (String methodName: methodNames) {
	try {
		Method method = obj.getClass().getMethod(oldMethodName);
		return (R) method.invoke(obj);
	} catch (Error e) { // try next one
	}
}
throw new IllegalStateException("Unable to call " + methodNames[0] + "on " + obj ,e);
```
Just coded this in the browser so this might need some formatting.

HannesWell (approved):
That should work for all cases!
Thanks.

Just a little alternative suggestion from me that needs the reflection code only once. But either way should work.

PR #1390, opened 2022-10-10
[eclipse/xtext#1982] support for java 17

cdietrich:
[eclipse/xtext#1982] support for java 17


org.eclipse.xtend.core/src/org/eclipse/xtend/core/javaconverter/JavaASTFlattener.xtend
```diff
@@ -1774,7 +1774,14 @@ class JavaASTFlattener extends ASTVisitor {
 			appendToBuffer("default ")
 		} else {
 			appendToBuffer("case ")
-			node.getExpression().accept(this)
+			if (ASTParserFactory.asJLS(targetApiLevel) >= 14) {
```
szarnekow, line 1777: IIRC this is J2X - shall we drop that feature?

szarnekow (approved):

org.eclipse.xtend.core/src/org/eclipse/xtend/core/javaconverter/JavaASTFlattener.xtend
cdietrich: think about. yes

PR #1416, opened 2022-11-25
Remove Dependency to deprecated junit plugins

cdietrich:
Remove Dependency to deprecated junit plugins


org.eclipse.xtend.standalone/META-INF/MANIFEST.MF
```diff
@@ -5,9 +5,7 @@ Bundle-Name: Xtend Standalone Bundle
 Bundle-Vendor: Eclipse Xtext
 Bundle-Version: 2.30.0.qualifier
 Require-Bundle: org.eclipse.xtend.core;visibility:=reexport,
- org.eclipse.xtext.junit4;visibility:=reexport,
```
cdietrich, line 8: @szarnekow do you know what this plugin is good for at all?

szarnekow: Might be that the on-the-fly compiler was used from there.

szarnekow (approved):

PR #1426, opened 2022-12-08
[eclipse/xtext#2125] set minimum junit version to 4.13.2

cdietrich:
[eclipse/xtext#2125] set minimum junit version to 4.13.2


szarnekow:
Failing test is unrelated as it seems?

PR #1441, opened 2023-01-12
fix Default/Instance/Configuration Scope deprecations

cdietrich:
fix Default/Instance/Configuration Scope deprecations


szarnekow (approved):
Getting rid of deprecations is always nice. Thank you for cleaning this up.

Issue #149, opened 2017-02-03
Unclear error for extension providers on interfaces


sbosselmann:
I was wondering whether extension providers work on interfaces so I gave it a try.

```java
interface MyInterface {
	final extension MyExtension = new MyExtension
}
```
It causes an error. The keyword `extension` is annotated with:

`Illegal modifier for the field null; only public, static, final & val are permitted`

However, the generated Java code is fine and makes it through the compiler with no issues. 
```java
public interface MyInterface {
  @Extension
  public final static MyExtension _myExtension = new MyExtension();
}
```

As I did not find anything on this issue here nor over at Bugzilla, I am still wondering whether Extension Providers should work on Interfaces or not. In the latter case, the error message really should be more meaningful.

kthoms:
Actually it does not make sense to have extensions in interfaces. Extensions are adding behaviour, and interfaces do not contain behaviour. Interfaces might contain constant values, method declarations or classes.

I consider that error as works-as-designed.

> Am 03.02.2017 um 18:13 schrieb sbosselmann <notifications@github.com>:
> 
> I was wondering whether extension providers work on interfaces so I gave it a try.
> 
> interface MyInterface {
> 	final extension MyExtension = new MyExtension
> }
> It causes an error. The keyword extension is annotated with:
> 
> Illegal modifier for the field null; only public, static, final & val are permitted
> 
> However, the generated Java code is fine and makes it through the compiler with no issues.
> 
> public interface MyInterface {
>   @Extension
>   public final static MyExtension _myExtension = new MyExtension();
> }
> As I did not find anything on this issue here nor over at Bugzilla, I am still wondering whether Extension Providers should work on Interfaces or not. In the latter case, the error message really should be more meaningful.
>

juangamnik:
> Actually it does not make sense to have extensions in interfaces. Extensions are adding behaviour, and interfaces do not contain behaviour. Interfaces might contain constant values, method declarations or classes.
> 
> I consider that error as works-as-designed.

Since Java 8 interfaces can have default methods and therefore they can clearly provide behavior. But aside from this, interfaces are the only types in Java (and Xtend) that allow (some kind of) multiple inheritance. 

Therefore, if you would like to add a common extension provider for all classes of "some kind" you now have to use an (abstract) parent class for this. But if you would like to use this super class w.l.o.g. for let's say a subclass of `JFrame`, you cannot extend at the same time the class with the extension providers.

I consider that error as an error.

[...]

PR #1444, opened 2023-01-20
[#149] allow extension fields in java 8 interfaces

cdietrich:
[#149] allow extension fields in java 8 interfaces

### DIFF
--- a/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/java8/compiler/Java8Compiler2Test.xtend
+++ b/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/java8/compiler/Java8Compiler2Test.xtend
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2020 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2015, 2023 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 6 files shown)


szarnekow:
I assume the extensions are available in classes implementing the interface, too

szarnekow (approved):
Nice

cdietrich:
> I assume the extensions are available in classes implementing the interface, too 
didnt test this. do we look up extensions in superclasses?

szarnekow:
I think we consider inherited extensions. It should just work

cdietrich:
it does not.  there is this done?

szarnekow:
> it does not. there is this done?

Hmm for classes I see the equivalent example working fine:
```
class A {
	public static extension String s;
	def void m1(int i) {
		i.substring();
	}
}
class B extends A {
	def void m2(int i) {
		i.substring();
	}
}
```

szarnekow:
And indeed it fails for interfaces:
```
interface A {
	@Extension
	public static String s;
	def void m1(int i) {
		i.substring();
	}
}
class B implements A {
	def void m2(int i) {
		i.substring(); // error
	}
}
```

szarnekow:
`org.eclipse.xtext.xbase.typesystem.internal.LogicalContainerAwareReentrantTypeResolver.addExtensionFieldsToMemberSession(ResolvedTypes, IFeatureScopeSession, JvmDeclaredType, JvmIdentifiableElement, Set<String>, Set<JvmType>)` is only traversing the extended class with the wrong assumption that interfaces don't define fields.

cdietrich:
replaced with https://github.com/eclipse/xtext/pull/2591

Issue #1445, opened 2023-01-23
Move to Maven/Tycho for projects and build infrastructure


LorenzoBettini:
See also https://github.com/eclipse/xtext-lib/issues/499 and https://github.com/eclipse/xtext-core/issues/2052 https://github.com/eclipse/xtext-extras/issues/858

Development has started in my fork's branch:
https://github.com/LorenzoBettini/xtext-xtend/tree/lb_2052_maven_tycho

LorenzoBettini:
@cdietrich I seem to understand that we must use Java 17, because

```
Error:  org.eclipse.m2e.core 2.0.5.20221128-1757 requires Execution Environment that matches (&(osgi.ee=JavaSE)(version=17)) but the current resolution context uses [a.jre.javase 11.0.0]
Error:  Cannot resolve target definition:
Error:    Software being installed: org.eclipse.m2e.feature.feature.group 2.1.2.20221130-2239
Error:    Missing requirement: org.eclipse.m2e.core 2.0.5.20221128-1757 requires 'osgi.ee; (&(osgi.ee=JavaSE)(version=17))' but it could not be found
Error:    Cannot satisfy dependency: org.eclipse.m2e.feature.feature.group 2.1.2.20221130-2239 depends on: org.eclipse.equinox.p2.iu; org.eclipse.m2e.core [2.0.5.20221128-1757,2.0.5.20221128-1757]
```

cdietrich:
i dont get this point
- r202203 target => java 11
- r202206 and later => java 17

LorenzoBettini:
Sorry, I can't see the current configuration of Jenkins jobs due to the outage, and up to now, I thought we were building with Java 11 (that's what I did in GitHub Actions). Since in the other projects, m2e wasn't being used, I realized that only now. My mistake :)

cdietrich:
nightly tiggered from xtext-eclipse https://github.com/eclipse/xtext-eclipse/blob/ef670c7566a58d3b75337660708d7c7d661e8e20/Jenkinsfile#L27

LorenzoBettini:
@cdietrich @szarnekow when running in GitHub Actions, this particular test fails on Windows (works on macOS and Linux):

`org.eclipse.xtend.core.tests.compiler.batch.BatchCompilerTest.testCompileSymlinkedResource()`

```java
	@Test
    def void testCompileSymlinkedResource() {
[...]

PR #1446, opened 2023-01-26
 Move to Maven/Tycho for projects and build infrastructure

LorenzoBettini:
See #1445 

Lots of detailed notes/questions/todos:

- xtend-maven-plugin is configured in pluginManagement and activated on single projects when needed (I did that also on the other Git repositories).
- there were two aggregator POMs tycho-pom-xml and maven-pom.xml. I renamed on the root folder tycho-pom.xml to pom.xml and I created another aggregator pom based on maven-pom.xml in a new project releng/org.eclipse.xtend.maven.releng. The new aggregator POM for Maven also has xtend.core as a child module, so that it can be run in isolation and does not assume that xtend.core has already been packaged (see also below the dedicated part for the Maven plugin).
- in the parent POM tycho-packaging-plugin as in other Git repositories I added
```
<!-- required to have valid consumable Maven artifacts
https://github.com/eclipse/tycho/issues/781 -->
<skipPomGeneration>true</skipPomGeneration>
```
- this profile is not "active by default" anymore, even because "activeByDefault" does not behave as expected (https://www.lorenzobettini.it/2022/12/maven-profiles-really-active-by-default/) and should never be used:
```xml
<profile>
	<id>r202203</id>
	<properties>
		<tycho-version>2.6.0</tycho-version>
		<target-platform-classifier>org.eclipse.xtend.target-r202203</target-platform-classifier>
	</properties>
</profile>
```
- xtend.core
additional.bundles added org.eclipse.emf.mwe2.lib
- xtend.core.tests
modified
xtend.core.tests.fast (xtend).launch
adding
```xml
<stringAttribute key="org.eclipse.jdt.launching.VM_ARGUMENTS" value="-Xmx1500m -ea"/>
```
or we get OutOfMemory error in some performance tests
removed launches xtend.core.tests.slow (xtend) and xtend.core.tests.slow because longrunning/src and longrunning/xtend-gen don't exist.
Renamed following convention a few Java classes:
```
SmokeTestCases -> SmokeTestCase
Case_14 -> Case_14_Test
TestBatchCompiler -> BatchCompilerTest
Java8CompilerTest2 -> Java8Compiler2Test
RewritableImportSectionTest0
RewritableImportSectionTest1
RewritableImportSectionTest2
TestProperSourceFolderMapping
Java8CompilerTest1
TryWithResourcesTestJava8
RichStringCompilerTest2
```
The smoke suite not executed by the build, right?
- xtend.ide is not deployed on Maven central, right?
- target platform configuration:
referring to a target file instead of the maven project allows us to build single project without first installing the maven artifact of the target platform
```xml
<plugin>
	<groupId>org.eclipse.tycho</groupId>
	<artifactId>target-platform-configuration</artifactId>
	<version>${tycho-version}</version>
	<configuration>
		<target>
			<file>
				${project.basedir}/../releng/org.eclipse.xtend.target/${target-platform-classifier}.target
			</file>
		</target>
...
```
- xtend.ide.tests
To be able to run UI tests on macOS m1, I had to add this in the tests parent for macOS profile
```xml
<plugin>
	<groupId>org.eclipse.tycho</groupId>
	<artifactId>target-platform-configuration</artifactId>
	<configuration>
		<environments combine.children="append">
			<environment>
				<os>macosx</os>
				<ws>cocoa</ws>
				<arch>aarch64</arch>
			</environment>
		</environments>
...
```
I did not put in the main parent and in the main target platform configuration because support for m1 has been added only in the latest Eclipse versions.
In Jenkins we run Linux only so it won't be a problem.
As usual, renamed a few test cases to match surefire conventions.
There's another important issue in xtend.ide.tests: `org.eclipse.xtend.ide.tests.compiler.batch.XtendCompilerAntTaskTest` needs dependency JARS to be available in `target/antDeps`.
I reuse the configuration already available in the POM, that copies the JARs in that directory.
In fact, it was working when running the build.
The problem is that one has to know about that, and run at least `mvn package` once, before running those tests from Eclipse.
I left in the root folder a launch configuration `RUN_ME_before_tests.launch`. This requires some time because it runs the build (without tests), downloading lots of things from Maven Central and p2 repositories.
For this reason, running that from Oomph might be overwhelming.
But the "problem" was like that in the past as well.
Depending on the intended use-case, we might also move the `org.eclipse.xtend.ide.tests.compiler.batch.XtendCompilerAntTaskTest` in a separate test project.
The copying of the JARs is needed only this specific test case, if I understand correctly.
- xtend.standalone:
This is not published anymore right?
On Maven central it's 9 years old (2.4.3).
Is it a still valid project?
I added
```
Automatic-Module-Name: org.eclipse.xtend.standalone
Bundle-RequiredExecutionEnvironment: JavaSE-11
```
and created a .classpath
- xtend.performance.tests
the launch configuration was missing
```
<booleanAttribute key="run_in_ui_thread" value="true"/>
```
so `org.eclipse.xtend.ide.tests.AbstractXtendUITestCase.tearDown()` was throwing an NPE.
I removed /org.eclipse.xtend.performance.tests/performance-test.cquery which was from the good ol' Buckminster (how many good memories ;)
- xtend.ide.swtbot.tests
Most of the tests are ignored. Is that intended?
Only this remains: `org.eclipse.xtend.ide.tests.preferences.FormatterPreferencesTest`.
This was failing when run from Eclipse from my mac m1.
Opening preferences dialog on mac did not work and it was too complex (maybe the m1 is too fast, and this code might lack synchronization?):
```java
/**
 * Opens the Eclipse Preferences dialog. On MacOS the preferences are in the system menu which is not accessible via SWTBot.
 * 
 */
public PreferencesDialogAPI openPreferences() {
	// see https://www.eclipse.org/forums/index.php/t/854280/
	if (Platform.getOS().equals(Platform.OS_MACOSX)) {
		// On Mac, the Preferences menu is under the system menu
		final IWorkbench workbench = PlatformUI.getWorkbench();
		workbench.getDisplay().asyncExec(() -> {
			IWorkbenchWindow window = workbench.getActiveWorkbenchWindow();
			if (window != null) {
				Menu appMenu = workbench.getDisplay().getSystemMenu();
				Optional<MenuItem> item = Arrays.stream(appMenu.getItems())
						.filter(i -> i.getText().startsWith("Preferences") || i.getText().startsWith("Einstellungen")).findFirst();
				if (item.isPresent()) {
					MenuItem menuItem = item.get();
					Event event = new Event();
					event.time = (int) System.currentTimeMillis();
					event.widget = menuItem;
					event.display = workbench.getDisplay();
					menuItem.setSelection(true);
					menuItem.notifyListeners(SWT.Selection, event);
				}
			}
		});
	} else {
		bot.menu("Window").menu("Preferences").click();
	}
	return new PreferencesDialogAPI(bot.shell("Preferences"));
}
```
In my experience, this is the correct and reliable way to do it and I confirm it works
```java
public PreferencesDialogAPI openPreferences() {
	// see https://www.eclipse.org/forums/index.php/t/854280/
	if (Platform.getOS().equals(Platform.OS_MACOSX)) {
		// On Mac, the Preferences menu is under the system menu
		// so we must open it programmatically
		UIThreadRunnable.asyncExec(() -> {
			PreferenceDialog dialog = PreferencesUtil
				.createPreferenceDialogOn
					(bot.activeShell().widget, null, null, null);
			dialog.open();
		});
	} else {
		bot.menu("Window").menu("Preferences").click();
	}
	return new PreferencesDialogAPI(bot.shell("Preferences"));
}
```
I don't think this additional configuration is needed:
```xml
<profile>
	<id>latest</id>
	<build>
		<plugins>
			<plugin>
				<groupId>org.eclipse.tycho</groupId>
				<artifactId>target-platform-configuration</artifactId>
				<configuration>
					<dependency-resolution>
						<extraRequirements>
							<requirement>
								<type>eclipse-plugin</type>
								<id>org.apache.felix.scr</id>
								<versionRange>1.0.0</versionRange>
							</requirement>
							<requirement>
								<type>eclipse-plugin</type>
								<id>org.eclipse.equinox.event</id>
								<versionRange>1.0.0</versionRange>
							</requirement>
							<requirement>
								<type>eclipse-plugin</type>
								<id>org.eclipse.pde.ui</id>
								<versionRange>1.0.0</versionRange>
							</requirement>
							<requirement>
								<type>eclipse-feature</type>
								<id>org.eclipse.platform</id>
								<versionRange>3.6</versionRange>
							</requirement>
							<requirement>
								<type>eclipse-feature</type>
								<id>org.eclipse.jdt</id>
								<versionRange>1.0.0</versionRange>
							</requirement>
							<requirement>
								<type>eclipse-plugin</type>
								<id>org.eclipse.xtext.common.types.shared.jdt38</id>
								<versionRange>1.0.0</versionRange>
							</requirement>
							<requirement>
								<type>eclipse-plugin</type>
								<id>org.eclipse.xtext.logging</id>
								<versionRange>0.0.0</versionRange>
							</requirement>
							<requirement>
								<type>eclipse-plugin</type>
								<id>org.slf4j.binding.simple</id>
								<versionRange>0.0.0</versionRange>
							</requirement>
						</extraRequirements>
					</dependency-resolution>
				</configuration>
			</plugin>
		</plugins>
	</build>
</profile>
```
Locally, I'm building with the latest, without activating the latest profile and everything works.
I haven't removed that, but @cdietrich do you remember why that's required?
- p2repository renamed (without '-') and moved one level up to be consistent with the other Git repositories

**Maven plugin:**

- xtend.maven.releng project added as said above, replacing maven-pom.xml
```xml
<!-- 
  Maven repository initially produced by gradle upstream build.
  Used to grab the upstream maven artifacts like org.eclipse.xtend.core
  and also to deploy the artifacts produced in this build.
 -->
<gradleMavenRepo>file:${root-dir}/build/maven-repository</gradleMavenRepo>
```
this should not be needed?
Note these dependencies (I haven't touched them) are correct once we're on the same Git repository or we use useJenkinsSnapshots:
```xml
<dependency>
	<groupId>org.eclipse.xtext</groupId>
	<artifactId>org.eclipse.xtext.xbase</artifactId>
	<version>${project.version}</version>
</dependency>
<dependency>
	<groupId>org.eclipse.xtext</groupId>
	<artifactId>org.eclipse.xtext.xbase.lib</artifactId>
	<version>${project.version}</version>
</dependency>
<dependency>
	<groupId>org.eclipse.xtext</groupId>
	<artifactId>org.eclipse.xtext.util</artifactId>
	<version>${project.version}</version>
</dependency>
<dependency>
	<groupId>org.eclipse.xtext</groupId>
	<artifactId>org.eclipse.xtext.smap</artifactId>
	<version>${project.version}</version>
</dependency>
<dependency>
	<groupId>org.eclipse.xtend</groupId>
	<artifactId>org.eclipse.xtend.lib</artifactId>
	<version>${project.version}</version>
</dependency>
```
I removed
```xml
<repository>
	<id>local-repo</id>
	<url>${gradleMavenRepo}</url>
</repository>
```
because we build all the required projects in the same reactor from scratch.
Note that xtend.core is built from the actual project (so Tycho kicks in for the xtend.core project which is right!) during the building of the projects of the Maven plugin.
I guess this should be removed as well:
```xml
	<distributionManagement>
		<repository>
			<id>local-repo</id>
			<url>${gradleMavenRepo}</url>
		</repository>
		<snapshotRepository>
			<id>local-repo</id>
			<url>${gradleMavenRepo}</url>
		</snapshotRepository>
	</distributionManagement>
```
- final note on the maven plugin: no tests are executed because they are excluded by default (it was already like that in the POM). Is that intentional?


cdietrich:
@LorenzoBettini wonder why there still is a maven build
how does this one find its dependencies?

LorenzoBettini:
> @LorenzoBettini wonder why there still is a maven build how does this one find its dependencies?

What are you referring to in particular? I mean, which point.

cdietrich:
https://github.com/eclipse/xtext-xtend/blob/f6ef3e9b8f707b781cdfa3c28902d3270b2a6810/Jenkinsfile#L81
here there still is a separate build. also the files are (not yet renamed order wise)
so it does not work to build these as part of the tycho build?

LorenzoBettini:
So you're asking how it could work in the past?

In the past, you first ran the Tycho build, specifying `install` so that artifacts are installed in the local .m2, and then you ran the Maven build without `xtend.core` in the reactor, but it was found in the local .m2.
Is that what you're asking?

There is still a separate build just to make it easier to deploy the Maven artifacts on Maven Central without explicitly disabling the other projects from deployment.
Of course, what I say makes sense if only the children of this POM https://github.com/eclipse/xtext-xtend/blob/lb_2052_maven_tycho/releng/org.eclipse.xtend.maven.releng/pom.xml are the ones to be deployed to Maven Central.
Moreover, if we merge Git repos we'll have to adjust and update all these things.

cdietrich:
dont we have to deploy core anyway as the plugin depends to it?

LorenzoBettini:
> dont we have to deploy core anyway as the plugin depends to it? i see we build core twice. is the artifact produed by both the same?

yes, we have to deploy it to Maven central, that's why now it's in the reactor.

In any case, it's better to deploy to Central and to p2 repositories, separately, artifacts that are NOT the same.

cdietrich:
hmmm question wopuld be how they differ?

LorenzoBettini:
artifacts in p2 repos typically have a build qualifier in the version, e.g., 2.30.0.v...dateandtime... while on central they are 2.30.0 (see e.g. https://central.sonatype.dev/artifact/org.eclipse.emf/org.eclipse.emf.ecore/2.29.0), though of course, you can also release them with a build qualifier as long as you turn SNAPSHOT into RELEASE. Besides that, they are typically the same. At the binary level, they're the same, since they are packaged by the same tool.

cdietrich:
but the content of jar is the same correct?

LorenzoBettini:
> but the content of jar is the same correct? this is what i was concerned

the .class files are the same. The MANIFEST and other metadata like the generated pom inside the META directory differ for the version as I said above.

cdietrich:
with the repo merge done this is obsolete

